PR #9, opened 2020-03-07
Initialize node

slifty:
## Description
This PR adds a package.json and eslint file

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits
- [x] Relevant labels assigned

## Steps to Test
none

## Deploy Notes
- `yarn install`

## Related Issues
Related to #6

Issue #6, opened 2020-02-24
Set up Node stack


slifty:
It should eventually include:

- [x] Jest
- [ ] ~Sequelize + psql (`pg`)~
- [x] TravisCI (#4 captures this already)
- [ ] ~PM2 for production (or an alternative)~
- [x] Linting + transpiling (ESLint + Babel)
- [x] Editorconfig
- [ ] Winston for logging

## Things to consider:

This will likely be for a future issue but documenting them here.

* GraphQL as an API option
* Kafka for coordination between modules.

slifty:
I'm thinking we should hold off on having sequelize and PM2 as part of the "node stack" -- lets add these if / when we decide to use them.

I am going to add winston however.

PR #10, opened 2020-03-07
Add jest

slifty:
## Description
This PR adds support for writing tests (`jest` and `nock`).

It also creates a new yarn command `yarn test` which will run automated tests.

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits

## Steps to Test
1. `yarn test`

(This will complain about there being no tests.)

## Deploy Notes
- `yarn install` 

## Related Issues
Related to #6 
Relies on #9


src/tests/hello.test.js
```diff
@@ -0,0 +1,6 @@
+
+describe('A hello world test that we will delete', () => {
```
reefdog: Hey we could delete it, _or_ we could always leave this in as a test of the testing system itself!

slifty: Ok that's a fair point -- renamed things so that would be the case.

PR #11, opened 2020-03-08
Add eslint

slifty:
## Description
This PR adds ESLint to our project, which we're using to make sure styles are followed consistently.

It also adds related plugins (airbnb) and eslint configurations.  The only changes from airbnb default are:

1. No semicolons
2. Tabs instead of spaces

Finally, this adds a new yarn script (`yarn lint`).

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits

## Steps to Test
1. `yarn lint`

## Deploy Notes
- `yarn install`

## Related Issues
Related to #6 
Relies on #10

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,0 +1,18 @@
+{
+	"extends": "airbnb",
+	"rules": {
+		"semi": [2, "never"],
+		"indent": [2, "tab"],
+		"no-tabs": 0
+	},
+	"env": {
+		"es6": true,
+		"node": true,
+		"jest": true
+	},
+	"settings": {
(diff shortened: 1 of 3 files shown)


slifty:
It looks like you might have a global eslint configured that is causing problems (... I did not know this was a thing that could happen!)

reefdog:
You are correct and I have no idea why! Maybe VS Code created that? But yes, I had it at `~/Projects/.eslintrc` and removing that let `yarn lint` run w/out errors. That's… weird and alarming! But means this is g2g!

PR #12, opened 2020-03-08
Add babel

slifty:
## Description
This PR adds babel for transpiling, and configures it as well. It modifies our linting configuration to properly work with babel.

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits

## Steps to Test
1. `yarn start`

## Deploy Notes
- `yarn install`

## Related Issues
Related to #6 
Relies on #11

Related to {Y}

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,15 @@
+{
+	"presets": [
+		[
+			"@babel/preset-env",
+			{
+				"targets": {
+					"node": "13.10"
+				}
+			}
+		]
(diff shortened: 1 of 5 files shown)


slifty:
We spoke by voice about this after the comment -- one concern with `.nvmrc` is that it is a developer specific tool.  Not everybody will use nvm for version management.

As a compromise we are going to document what to do if you get the error.

PR #18, opened 2020-03-16
Use Markdown headers for issue template sections

reefdog:
This replaces the simple bolded text with `####` (H4) headers, which render in a similar size w/in GitHub but add consistent margin between the header and the following content. (Otherwise, whether the next line of text appeared directly below the header or not would depend on if the contributor entered a newline between them or not, and that’s just sloppy.)

It also adds a much-needed period to one of the descriptions.

For some reason, opening this PR didn't prefill with the PR template. But in any case, it's just a GitHub-specific issue template change with no associated issue.

### DIFF
--- a/.github/ISSUE_TEMPLATE/discussion.md
+++ b/.github/ISSUE_TEMPLATE/discussion.md
@@ -9,8 +9,8 @@ assignees: ''
 
 ## Discussion
 
-**What do you want to talk about?**
+#### What do you want to talk about?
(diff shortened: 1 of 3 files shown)


slifty:
While you're editing these -- you wanted to add a "." to the end of one of the issue templates.  May as well sneak that into this PR?

reefdog:
Ah yes. Done!

PR #32, opened 2020-04-14
Create project structure

### LINKED ISSUE #31: Let's talk directory structure
#### What do you want to talk about?
Right now we have:

* `/docs`: documentation and such
* `/services`: dockerized containers for various external tools we will rely on (Kafka, but eventually sql I imagine)
* `/src`: where the code lives

The question is how do we want to organize the code inside of `/src`.

There are tons of articles about node code structure, though many are assuming a react application which isn't really this.

slifty:
## Description
This PR adds a more reasonable directory structure to the project overall.

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits

## Steps to Test
1. `yarn test`

## Deploy Notes
None

## Related Issues
Resolves #31


README.md
```diff
@@ -24,6 +24,23 @@ yarn kafka:start
...
+```
+- docs     // Project documentation
+- services // Convenience containers for various external services (e.g. Kafka)
+- src      // The primary code location itself
+| - components // Architectural components
+| - constants  // Constants used throughout the project
+| - lib        // Supporting code (e.g. utilities, DB singletons, etc)
+| - scripts    // Supporting scripts
+| - tests      // System-wide / integration tests
+```
```
reefdog, line 42: Are we going to have a `/src/models`, or were you thinking that would be in `/src/lib/models`? We don't need to mock it out in advance, just curious.

README.md
```diff
@@ -24,6 +24,23 @@ yarn kafka:start
...
 
+## Project Structure
+
+To understand the overall architecture of the project, please review our [architecture documentation](docs/ARCHITECTURE.md).
+
+The root structure is as follows:
+
+```
+- docs     // Project documentation
+- services // Convenience containers for various external services (e.g. Kafka)
```
reefdog, line 35: `/services` confuses me a little bit, but I think it'll make sense as more stuff develops.

README.md
slifty: Open to talking through it when the time comes, but I think it should be `/src/models`

README.md
slifty: The logic here is those really are (containerized) *services* -- kafka, postgres.

Maybe `/serviceContainers` is a better name, but I also didn't want to assume everything will be in a container necessarily.

We can definitely tweak over time.  The stuff in that folder is not going to be directly referenced by anything in `src` so it should be easy to modify.

PR #36, opened 2020-04-24
Configure Jest to support absolute imports

### LINKED ISSUE #35: Jest chokes on our absolute imports
#### Description
Out of the box, Jest does not support the `module-alias`-mediated absolute imports. When you try to run tests on code that eventually resolves to one of these absolute imports, it doesn't know what to do.

Happily, this is [known and has a resolution](https://github.com/ilearnio/module-alias#usage-with-jest): [Jest's `moduleNameMapper` configuration](https://jestjs.io/docs/en/configuration#modulenamemapper-objectstring-string--arraystring).

reefdog:
## Description
We use `module-alias` to let us absolutely import modules from defined roots. However, Jest doesn’t pick up these defined roots automatically; instead, we need to replicate them in our Jest configuration.

This PR adds that configuration. Note that Jest uses regex and suggests you include beginning/end anchors for explicitness, so we do.

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits

## Steps to Test
1. `yarn test`

## Deploy Notes
None.

## Related Issues
Resolves #35

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,12 +1,5 @@
-const package = require('./package.json')
+const { getModuleAliasesForESLint } = require('./lib/utils/moduleAlias')
 
-const extractModuleAliasesFromPackage = (package) => {
-	const aliasStems = Object.keys(package._moduleAliases)
-	return aliasStems.map(
-		(stem) => [stem, package._moduleAliases[stem]]
-	)
-}
-const aliases = extractModuleAliasesFromPackage(package)
 const settings = {
 	"extends": "airbnb-base",
 	"parser": "babel-eslint",
@@ -22,7 +15,7 @@ const settings = {
 	},
 	"settings": {
 		"import/resolver": {
-			"alias": aliases
+			"alias": getModuleAliasesForESLint()
 		}
 	}
 }
--- a/lib/utils/moduleAlias.js
+++ b/lib/utils/moduleAlias.js
@@ -1,0 +1,21 @@
+// Require our actual `package.json` data for use by this file.
+//
+// NB: `package` is a reserved word
+const packageData = require('../../package.json')
+
+// Extracts and returns the `module-alias` configuration from our `package.json` data.
+//
+// We don't control the `_moduleAliases` variable name, we just work with it.
+// eslint-disable-next-line no-underscore-dangle
+const getModuleAliasesFromPackage = () => packageData._moduleAliases || {}
+
+// Loads aliases from our `package.json` data and returns them, transformed into ESLint's
+// configuration format: `[[alias, destination]]`.
+const getModuleAliasesForESLint = () => {
+	const moduleAliases = getModuleAliasesFromPackage()
+	return Object.keys(moduleAliases).map((stem) => [stem, moduleAliases[stem]])
+}
+
+module.exports = {
+	getModuleAliasesForESLint,
+}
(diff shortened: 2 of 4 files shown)


slifty:
Looks reasonable!

One thing for consideration: Jest [lets you configure](https://jestjs.io/docs/en/configuration) via a `jest.config.js` file.  Since we're configuring our packages in config files for the most part, it feels maybe more explicit / clean to do the configuration there instead of in our `package.json`

Another, related thought is that our `.eslintrc.js` file does a similar thing for configuring eslint to use our aliases.  I wonder if we could follow a similar pattern of importing the package.json and constructing the jest-ified version of the module aliases.

This would mean we no longer have to remember to update two things if we add a new alias.

reefdog:
@slifty Okay! I implemented the suggestion to create a central shared tooling utility for extracting `package.json`'s `module-alias` configuration and transforming it to a tool-specific configuration, then using that for ESLint and Jest. Let me know what you think…

.eslintrc.js
```diff
@@ -22,7 +15,7 @@ const settings = {
 	},
 	"settings": {
 		"import/resolver": {
-			"alias": aliases
+			"alias": getModuleAliasesForESLint()
```
slifty, line 18: this is sooooooo much nicer

lib/utils/moduleAlias.js
```diff
@@ -0,0 +1,38 @@
+// Require our actual `package.json` data for use by this file.
+//
+// NB: `package` is a reserved word
+const packageData = require('../../package.json')
+
+// Extracts and returns the `module-alias` configuration from our `package.json` data.
```
slifty, line 6: Love the documentation, but we should use the docblock format that we've been using elsewhere.  I believe that adheres to [this standard](https://jsdoc.app/).

(side note: I am using [DocBlockr](https://packagecontrol.io/packages/DocBlockr) for sublime, idk if there is a similar one for your IDE)

Key elements are using /** and */ as well as the use of things like `@returns` etc.

lib/utils/moduleAlias.js
```diff
@@ -0,0 +1,38 @@
...
+// Loads aliases from our `package.json` data and returns them, transformed into ESLint's
+// configuration format: `[[alias, destination]]`.
+const getModuleAliasesForESLint = () => {
+	const moduleAliases = getModuleAliasesFromPackage()
+	return Object.keys(moduleAliases).map((stem) => [stem, moduleAliases[stem]])
+}
+
+// Loads aliases from our `package.json` data and returns them, transformed into Jest's
+// configuration format: `{ alias: destination }`.
+const getModuleAliasesForJest = () => {
```
slifty, line 21: I just want to note that deep down I know we both really want there to be tests for these but I appreciate that WE MUST SACRIFICE THAT CONCEPT IN THE NAME OF PROGRESS.

reefdog: Yeah… I actually wasn't blocked on that by progress but by the fact that this function doesn't take input (intentionally), and instead relies on pulling live `package.json` data from the outer scope (again, by design). I… wasn't sure how to test that, and didn't want to spin wheels to figure it out (yes inthenameofprogress).

Feel free to open an issue, although, I also think we'd be made aware of any failures here quickly because linting and/or testing would break.

reefdog:
Oops, just asked for a re-review, but "no need for re-review". Going!

Issue #14, opened 2020-03-16
ffmpeg / stream exploration


slifty:
## Discussion

**What do you want to talk about?**

As we begin to work on the input coordinator / ingestion engines, we discussed that the primary purpose of these modules will be to convert the input into a standardized video stream type.

The question is what should that stream type be.  We should also think through what the likely input formats will be (file, youtube, youtube live, HD HomeRun)

**Relevant resources**

1. We have some access to folks who have worked a lot in this space, so lets tap into their recommendations.  @traceypooh in particular!
2. The [developer documentation of HDHomeRun](https://www.silicondust.com/hdhomerun/hdhomerun_development.pdf)
3. [ffmpeg formats](https://ffmpeg.org/ffmpeg-formats.html)
4. [ffmpeg streaming guide](https://trac.ffmpeg.org/wiki/StreamingGuide)

reefdog:
A few thoughts about standardizing the stream.

1. The easiest ingestion processing is no processing, so wherever possible, **configure the source input delivery format to match our standardized output preferences.** Meaning, if we decide to deal exclusively with h264 inside the kitchen, request content in h264 from our input streams (HDHomeRun, YouTube) so the ingestion engine doesn't have to convert. Similarly, if we want to deal internally with no more than 1080p video, request 1080p video sources so we're downscaling as rarely as possible.
2. **Minimize the number of format options we choose to standardize.** Do we need to standardize the frame rate, or can we commit to dealing internally with the native one? I can imagine setting a cap (like 30fps) beyond which we don't gain value, but below that cap, if we can pass through 24 or 25 or 29.97 or 30 unmodified, we again reduce our processing requirements.

_Sending this comment mid-flight even though I can/need to add more detail!_

slifty:
I think we can close this at this point!

We locked in on MPEG-TS as a container format, we also have a pretty robust ffmpeg stream pipeline about to be merged in as part of the work on #45.

I'm sure some day we'll want to do more exploration about optimizing encoding formats or some such thing, but for now I don't think we have active open questions.

PR #41, opened 2020-05-01
Demonstrate ffmpeg proxying of external to internal stream

reefdog:
## Description
This PR adds FFmpeg to our tech stack and demonstrates a very simple script that takes an input address (file or network stream) and proxies it through FFmpeg to an available internal network stream, configurable via `.env`.

To accomplish that though, it adds a bunch (possibly too much) of support infrastructure.

## Due Diligence Checklist
- [x] Tests
- [x] Documentation: TODO: Needs JSDoc format for tests
- [x] Consolidated commits

## Steps to Test
1. Modify `.env` to add the new variables from `.env.template`
2. `yarn ingest:showInternalStreamPool` should list out all the address/port combinations from ☝️
3. `yarn ingest {source}` should open a stream to the first port from that list; for now, a good `{source}` would be an HDHomeRun stream address like `http://<device ip>:5004/auto/v<channel number>`
4. To verify ☝️ is happening, open a new tab and run `ffplay {address}` where `{address}` is the first address dumped out by `yarn ingest:showInternalStreamPool` (which is the one it will be streaming to)

## Deploy Notes
1. Must modify `.env` to add the new variables from `.env.template`
2. Must install FFmpeg; this is outside of the scope of the instructions.

## Related Issues
Related to #14

### DIFF
--- a/.env.template
+++ b/.env.template
@@ -4,3 +4,16 @@ LOG_LEVEL=info
 
 # A comma-delimited list of Kafka brokers
 KAFKA_BROKERS=localhost:9092
+
+# The address and port range pool for the internal streams. The ingester will compose an address
+# from this pool for streaming its output to the countertop. Note that number of available ports is
+# effectively the cap on number of simultaneous streams.
+#
+# Set address to an IP address or `localhost`:
+INTERNAL_STREAM_ADDRESS=localhost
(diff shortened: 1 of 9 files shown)


reefdog:
We've decided to have the ingesters stream data into Kafka rather than via internal UDP streams, making this whole branch/exploration unneeded. :tada:

PR #43, opened 2020-05-16
Implement abstract ingestion engine

### LINKED ISSUE #15: Create the abstract ingestion engine
## Task

**Description**
The architecture outlined in #1 talks about the idea of a "ingestion engines" which will be responsible for converting external media sources into a standardized video stream for internal use.

We anticipate a few types of inputs over time, so we should create an abstract class that defines the general shape of what an "Ingestion Engine" is responsible for.

Location of this TBD but maybe put it in a new directory like `src/ingestionEngines/`?

**Relevant resources / research**
* Issue #1 has some architecture diagrams (which will eventually be put into the documentation repository)

**Related Issues**
* #14 Is an issue for the exploration of ffmpeg and stream formats that we might want to use

slifty:
## Description
**NOTE:** THIS IS A WIP BRANCH RIGHT NOW.  We need more tests before this is really ready to rock.

**NOTE 2:** This will fail CI, because `@tvkitchen/base*` is not yet published to npm.

This PR Adds the AbstractIngestionEngine, which implements the conversion of a ReadableStream of video into kafka.

It is possible that `STREAM.DATA` should be renamed to `VIDEO.STREAM_DATA` or `VIDEO_STREAM.DATA` or `STREAM.VIDEO_DATA` to eventually allow for other types of media streams (e.g. audio).

## Due Diligence Checklist
- [ ] Tests
- [ ] Documentation
- [x] Consolidated commits

## Steps to Test
1. `yarn install`
2. `yarn test`

## Deploy Notes
None

## Related Issues
Resolves #15


slifty:
@reefdog I'm not sure that things are baked enough to warrant the time testing *everything*

I've added a few tests, and left stubs for others.  I think this might be ready for review, but with a keen eye towards "actually you should write tests for this part".

slifty:
The woes of not writing tests.

1. getInputstream should return null.
2. we don't have a getCodecSettings or whatever

slifty:
JUST A NOTE: This PR is now a bit obsolete since #47 implements a few changes to the way the abstract ingestion engine works; that said, I think we should still review and merge this, just without requiring tests for the methods that got removed almost right away.

slifty:
Merging this weird PR which represents a moment in time and transition for AbstractIngestionEngine...

Issue #42, opened 2020-05-15
Determine how to track and keep time


reefdog:
## Discussion

"Time" in the context of the TV Kitchen has two meanings; I'll describe them both, but the thrust of this issue is the second one.

### 1. Determining when the content aired.

Airtime will be important downstream when users want to do time-based analysis like "how often was the economy discussed last month?" Our two working limiting assumptions are (a) we can only provide airtime for live streams, and (b) we'll generate airtime based on when we saw the live stream rather than being able to rely on stream-provided airtime metadata.

If we can do better than either of those two assumptions — such as if users overwhelmingly want us to treat the file creation/modification dates as local file airtimes, or if live streams bake in universal timecodes that are more accurate than our "when we saw it" proxy value — then we will.

I don't think this needs much discussion at this point. Our limiting assumptions mean the decisions are fairly straightforward.

### 2. Keeping audio/video/caption streams in sync throughout the pipeline.

This isn't necessary while we're still dealing with a container like MP4 or MPEG-TS, which are built to keep these streams in sync. But once we split the streams for processing by individual appliances, we'll need to use timecoding to be able to communicate that this frame appeared at the same time as this caption and with this speaker's voice talking, etc.

**There are significant open questions here, such as:**

1. Whose responsibility is it to keep time? Does the ingestion engine generate timecodes (or elevate internal ones) during ingestion, or is that the role of the Sous Chef? Or a "timekeeper" appliance that precedes all other steps?
2. How do we expose this timecode to the pipeline? Is it part of the [`Payload`](https://github.com/tvkitchen/base/blob/master/src/Payload.js) metadata, or is it buried in the `data` bitstream itself? Does this decision change based on the composition of the `data`? (Like, if the payload is carrying `TEXT.CHARACTER` instead of a packaged MPEG-TS stream, how _could_ we decorate the character data with a timecode _without_ using the Payload metadata attributes?)
3. Does the size of an individual `Payload` affect this decision? For instance, if a `Payload`'s `data` is a single character, can we even timecode granularly enough for "timecode" to have any relevance? (It already argues for not reconstructing a stream based on timecode along; we need another way of sequencing linear data streams, but I assume that'll be handled naturally by the Kafka queue orders.)
4. Do we need separate concepts for "timecode" (used for keeping a/v/c streams in sync) and "airtime", or can timecodes be expressed in UTC (or equivalent)?

slifty:
[...]

PR #47, opened 2020-05-27
Include presentation times in ingested container stream

### LINKED ISSUE #45: Track stream time
#### Description
Ingestion engines need to be able to insert time to payloads.

There is some discussion at #42 around handling the concept of time, and this can be done in parallel with that exploration.

The point of this issue is far more narrow: how do we know how much content has been processed so far from a given input.  (e.g. for a given payload of data added to kafka, how can we know how far into a stream or file that payload is linked to).

It seems ffmpeg will output "progress" on a regular cadence *to a different pipe than the data*  This is explored a bit in the discussion of Issue #44, and is probably going to be our best bet.

Another option would be to buffer the ffmpeg out and run it through ffprobe on a regular basis.  That feels... horrible!

None yet.

#### Related Issues
Issue #42 
Issue #44 is an exploration of libraries that can be used to track time

slifty:
## Description
* This PR adds a critical piece of metadata to data extracted from an ingestion stream and sent to kafka: time.

Specifically, it adds 'ptsTime' to the payload object, which is not yet formally supported by the Payload object in `@tvkitchen/base-classes` but gives us something to work with.

As you can see in issue #45, The amount of R&D that went into this PR is kind of wild, and it's thanks to the efforts of @reefdog @chriszs @traceypooh and @mhirsch that we managed to figure out the best way to do it.

* This PR also adds a more complete set of tests to AbstractIngestionEngine now that the functionality is a bit more defined, and constrains the way that we're integrating with ffmpeg which should have an added bonus of preventing some of the security concerns that was offered by passing a variable to `spawn`.

* This PR updates the AbstractIngestionEngine to use the most recent `STREAM.CONTAINER` dataType, as described in https://github.com/tvkitchen/base/issues/14

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits

## Steps to Test
1. `yarn install`
2. `yarn test`

## Deploy Notes
- `yarn install`

## Related Issues
Resolves #45 
Related to #42


src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -5,61 +5,90 @@ import {
...
 		}
 		this.producer = kafka.producer()
+		this.mpegtsDemuxer = new TSDemuxer(this.processDemuxedPacket)
+		this.ffmpegProcess = null
+		this.currentPts = 0
 	}
 
 	/**
-	 * Ingests data into the TV Kitchen pipeline.
+	 * Ingests a mpegts payload into the TV Kitchen pipeline.
```
reefdog, line 26: `an MPEG-TS payload`…

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -5,61 +5,90 @@ import {
...
-			readableStream.resume()
+	processDemuxedPacket = (packet) => {
+		this.currentPts = packet.pts
+	}
+
+	/**
+	 * Process a new chunk of data from an mpegts stream.
+	 * This data is used to update local state and ultimately pipe to kafka.
+	 *
+	 * @param  {Buffer} mpegtsData The latest arbitrary chunk of buffered mpegts data.
```
reefdog, line 58: I think it's worth clarifying that this is a chunk of arbitrary _size_, not, like, arbitrary composition or order.

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -5,61 +5,90 @@ import {
...
+	 */
+	processRawData = (rawData) => {
+		this.inputStream.pause()
+		this.ffmpegProcess.stdin.write(rawData)
+		this.inputStream.resume()
+	}
+
+	/**
+	 * Start the ingestion engine.
+	 * This kicks off the ffmpeg stream and input stream pipeline.
```
reefdog, line 85: I think this second sentence could be expanded just a bit to make the order of operations more clear and descriptive. For instance, we intentionally:

1. Spawn an FFMpeg process to listen for input on stdin, then
2. Start the input stream and pipe it to the now-listening FFMpeg process

As-is, the sequential order of that is implicit when it would be helpful for the description to make it explicit.

src/components/ingestion/AbstractIngestionEngine.js
slifty: I saw that when we were on-screen this afternoon and smirked at myself.

I swear I know how to use indefinite articles.

reefdog: You indefinitely don't.

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -5,61 +5,90 @@ import {
...
+	 * This data is used to update local state and ultimately pipe to kafka.
+	 *
+	 * @param  {Buffer} mpegtsData The latest arbitrary chunk of buffered mpegts data.
+	 */
+	processMpegtsData = (mpegtsData) => {
+		this.ffmpegProcess.stdout.pause()
+		this.mpegtsDemuxer.process(mpegtsData)
+		const payload = new Payload({
+			data: mpegtsData,
+			ptsTime: tsToSeconds(this.currentPts),
```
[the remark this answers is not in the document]
slifty: Argh is this exposing a code smell...

Would it be clearer if it was something more like:

```
const packet = this.getMostRecentPacket()
const payload = ... {
  ...
  ptsTime: tsToSeconts(packet.pts)
}
```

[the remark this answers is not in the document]
reefdog: …yes I think so! Because reading back over this… I think what I would _assume_ happens here is that `		this.mpegtsDemuxer.process(mpegtsData)` actually changes/transforms `mpegtsData` itself (even though that's JavaScript-impossible, I think, or at least _very_ unidiomatic), something that's necessary for when we pass it into the payload with `data: mpegtsData`. I think folks would be surprised to find out that `this.currentPts`, instead, is the recipient of `this.mpegtsDemuxer.process(mpegtsData)`.

So yeah, I think rewriting it closer to what you have it would improve legibility.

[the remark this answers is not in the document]
reefdog: _Oh, you edited with lots more info; now I have to re-read!_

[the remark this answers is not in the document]
chriszs: Hey there, I haven't spent enough time with this to know for sure, but using an on data event and pausing the stream here may not the most clean and efficient way to accomplish throttling. It's possible there's some limiting constraint here that I'm not seeing, or experience that informs this choice, so _waves hands_ ignore me if so, but stream backpressure might be a better way to do throttling. If you want, I can take a look at what's involved in that, and whether the result would be in fact cleaner/more efficient either for this PR or a subsequent one after creating an issue.

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -5,61 +5,90 @@ import {
...
+	}
+
+	/**
+	 * Process a new chunk of data from an mpegts stream.
+	 * This data is used to update local state and ultimately pipe to kafka.
+	 *
+	 * @param  {Buffer} mpegtsData The latest arbitrary chunk of buffered mpegts data.
+	 */
+	processMpegtsData = (mpegtsData) => {
+		this.ffmpegProcess.stdout.pause()
```
chriszs, line 61: Hey there, I haven't spent enough time with this to know for sure, but using an on data event and pausing the stream here may not the most clean and efficient way to accomplish throttling. It's possible there's some limiting constraint here that I'm not seeing, or experience that informs this choice, so waves hands ignore me if so, but stream backpressure might be a better way to do throttling. If you want, I can take a look at what's involved in that, and whether the result would be in fact cleaner/more efficient either for this PR or a subsequent one after creating an issue.

src/components/ingestion/AbstractIngestionEngine.js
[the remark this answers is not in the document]
slifty: @reefdog great point -- and also while reading your note I realized that instead of `getMostRecentPacket()` it might be more like `getMostRecentPacketMetadata()` or something along those lines.

Indeed, those are not actual "packets" from an mpegts spec perspective, but `Packet` as defined deep inside `TSDemuxer`

src/components/ingestion/AbstractIngestionEngine.js
slifty: Oooo nice catch -- and you actually just exposed a nasty bug!

So to your solid throttle point: I believe we no longer need (or want) to throttle in the way I was toying with during the R&D stuff -- where I was trying to make sure ffprobe and ffmpeg would process at roughly the same speed.

The *original* intent of the `.pause` and `.resume` was in order to prevent a race condition where kafka messages were sometimes being recorded out of order.

For instance:

```
- Data A arrives (new thread: A)
- A: Write to Kafka
- Data B arrives (new thread: B)
- B: Write to Kafka
- B: Write complete
- A: write complete
```

By pausing, we ensure that `Data B arrives` won't happen until after `A: Write complete`

However, this code *does not accomplish that goal*.  Specifically:

`this.ingestPayload(...)`

should be

`await this.ingestPayload()`

---

So, with that context I *think* we still need to pause / resume, as something of a mutex, or maybe what we really should do is use a *real* mutex plus the backpressure solution you're describing?

Did anything I just wrote actually make sense or did I accidentally just write 70 lines of buzzwords.

slifty: Side note: I have no idea how to reliably test for race conditions...

chriszs: You go to the track and drive there.

Okay, so I think that's in part because the events continue to fire even if a prior event hasn't completed. It's async with no way to wait. A through or write stream pattern would handle this more gracefully by preserving order without a mutex or pausing/resuming.

chriszs: Serious answer about how to test: you might mock out the Kafka ingest, then block on that until the second one fires.

slifty: Oh yes! I'll add a test like that.

And yes that's exactly it -- unless I paused, events would keep firing.

chriszs: Okay, expanding on this. What you've got now is essentially:
```js
ffmpeg.stdout.on('data',data => {
    ffmpeg.stdout.pause()
    ingestPayload(data)
    ffmpegs.stdout.resume()
})

inputStream.on('data',data => {
    inputStream.pause()
    ffmpegs.stdin.write(data)
    inputStream.resume()
})
```

What I'm talking about looks something like this:

```js
const processStream = Writable({
    write: (data, enc, done) => {
        ingestPayload(data)
            .then(response => done())
            .catch(done)
    }
});

ffmpeg.stdout.pipe(processStream)

inputStream.pipe(ffmpeg.stdin)
```

I took a stab at applying this to the PR branch in a fork, you can find that [here](https://github.com/tvkitchen/tv-kitchen/compare/45-parse-progress...chriszs:45-parse-progress). It looks this there's a way to suggest code changes to a set of lines in a PR comment, but this seems a little larger than that was designed for. Anyway, it's there as a reference/basis for future work if you want it.

One question I have is whether ts-demuxer's process method calls the method that updates the timestamp synchronously. If it doesn't, there may be some mismatch in timing. I read its [code](https://github.com/gliese1337/HLS.js/blob/master/demuxer/src/index.ts), but couldn't be 100% sure.

chriszs: Thinking about this further this morning, something like this is possible and a little clearer about what each step does:
```js
import { Writable, Transform } from 'stream'

const processStream = Transform({
	objectMode: true,
	transform: (data, enc, done) => {
		TSdemuxer.process(data)
		done(null, new Payload{ payload: data, pts: this.currentPts })
	},
});

const ingestStream = Writable({
	objectMode: true,
	write: (data, enc, done) => {
        	ingestPayload(data)
            		.then(() => done())
            		.catch(done)
	},
});

inputStream.pipe(ffmpeg.stdin)

ffmpeg.stdout.pipe(processStream).pipe(ingestStream)
```

Could alternatively use pipeline syntax:
```js
import { pipeline } from 'stream'

inputStream.pipe(ffmpeg.stdin)

pipeline(
	ffmpeg.stdout,
	processStream,
	ingestStream,
	() => console.log('done'),
);
```
Callback style shown but it's promisfiable.

slifty: Oh, this is so nice -- implementing now (you're already a co-author on this commit but I kinda wanna add your name twice)

slifty: This is now done -- and boy, testing this approach was not easy!

src/components/ingestion/AbstractIngestionEngine.js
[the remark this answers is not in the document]
slifty: Called this stuff `getMostRecentDemuxedPacket` which I think is the most accurate to what it is.

slifty:
A lot has been changed since the last reviews -- this is ready for a re-review!

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
 import { spawn } from 'child_process'
+import {
+	pipeline,
+	Writable,
+	Transform,
+} from 'stream'
```
chriszs, line 6: We should think about using minipass in the future. Here's [why](https://www.npmjs.com/package/minipass#differences-from-nodejs-streams).

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
+	 * packet demuxed from the mpegts stream.
+	 *
+	 * @return {Packet} The most recent Packet object extracted by TSDemuxer.
+	 */
+	getMostRecentDemuxedPacket = () => this.mostRecentDemuxedPacket
+
+	/**
+	 * Process a new chunk of data from an mpegts stream. The chunks passed to this
+	 * function should be presented sequentially, and combine to form a cohearant mpegts
+	 * data stream, but they can be of arbitray size.
```
chriszs, line 75:
```suggestion
	 * data stream, but they can be of arbitrary size.
```

slifty: Woah what even is this interface option...

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
+	 *
+	 * It then spawns the input stream as specified by the implementing class, and creates
+	 * a pipeline that passes input->ffmpeg->packaging->ingestion
+	 *
+	 * @return {Stream} the ingestion pipeline that just got started
 	 */
 	start = () => {
-		const ffmpegSettings = this.getFfmpegSettings()
-		const ffmpegProcess = spawn('ffmpeg', ffmpegSettings)
+		const ffmpeg = spawn('ffmpeg', this.getFfmpegSettings())
```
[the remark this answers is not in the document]
slifty: Great point.

It *shouldn't* cause testing hangs in `yarn test` style testing, since we mock spawn, spawn never actually gets called during the test suite.

Is it hanging during the tests / timing out?  Or something else.

traceypooh:
IF for some reason it'd be useful for me to make a small C-based binary that _both_ copies all packets (passthru) _and_ prints first PTS (or all PTS) to stderr, for a given MPEG-TS stream, as a single lightweight binary "one-shot" to reduce some complexity, potentially, LMK! :)

src/components/ingestion/AbstractIngestionEngine.js
chriszs: I found it when Googling “how to suggest changes to a pull request.” How to make them under item 6 [here](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request#adding-line-comments-to-a-pull-request) and how to incorporate them [here](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/incorporating-feedback-in-your-pull-request#applying-suggested-changes).

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
+import { tsToSeconds } from '%src/lib/utils/mpegts'
 
 class AbstractIngestionEngine {
 	constructor() {
 		if (this.constructor === AbstractIngestionEngine) {
 			throw new AbstractInstantiationError(this.constructor.name)
 		}
 		this.producer = kafka.producer()
+		this.mpegtsDemuxer = new TSDemuxer(this.processDemuxedPacket)
+		this.mostRecentDemuxedPacket = null
```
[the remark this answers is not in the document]
slifty: Definitely worth an issue around exploring (though that issue would need to be ice boxed until we get data flowing through the system).  It's also possible we could contribute to TSDemuxer to expose an API more suited to our needs.

src/components/ingestion/AbstractIngestionEngine.js
slifty: I'll spin up an issue for this as part of merging.

src/components/ingestion/AbstractIngestionEngine.js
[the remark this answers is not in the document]
chriszs: Yeah, sorry. Not in yarn test. I did my own ad hoc integration testing so I didn’t send you something broken.

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
+	 *
+	 * @param  {Packet} packet The latest TSDemuxer Packet object.
+	 */
+	processDemuxedPacket = (packet) => {
+		this.mostRecentDemuxedPacket = packet
 	}
 
 	/**
-	 * Continuously ingest of a stream of data.
+	 * Returns the most recent cohearant stream packet processed by the ingestion engine.
```
reefdog, line 64:
```suggestion
	 * Returns the most recent coherent stream packet processed by the ingestion engine.
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
+
+	/**
+	 * Updates ingestion state based on the latest demuxed packet data.
+	 *
+	 * This method is called by our mpegts demuxer, and allows the ingestion engine to track
+	 * the most recent demuxed packet.
+	 *
+	 * @param  {Packet} packet The latest TSDemuxer Packet object.
+	 */
+	processDemuxedPacket = (packet) => {
```
reefdog, line 59: 😬 Okay so feel free to discard this because we're bike-shedding now _but_ this isn't really _processing_ the packet, is it? I mean it's the callback that `TSDemuxer` calls when _it_ has processed the packet, and handed it back to you to do whatever with. So I feel like it should really be called something like `onPacketDemuxingComplete` or something? That would certainly help _me_ understand it with less cartography.

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
+	 * This packet is lower level than the mpegts container, and represents an audio or video
+	 * packet demuxed from the mpegts stream.
+	 *
+	 * @return {Packet} The most recent Packet object extracted by TSDemuxer.
+	 */
+	getMostRecentDemuxedPacket = () => this.mostRecentDemuxedPacket
+
+	/**
+	 * Process a new chunk of data from an mpegts stream. The chunks passed to this
+	 * function should be presented sequentially, and combine to form a cohearant mpegts
```
reefdog, line 74:
```suggestion
	 * function should be presented sequentially, and combine to form a coherent MPEG-TS
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
+	 * @param  {Packet} packet The latest TSDemuxer Packet object.
+	 */
+	processDemuxedPacket = (packet) => {
+		this.mostRecentDemuxedPacket = packet
 	}
 
 	/**
-	 * Continuously ingest of a stream of data.
+	 * Returns the most recent cohearant stream packet processed by the ingestion engine.
+	 * This packet is lower level than the mpegts container, and represents an audio or video
```
reefdog, line 65:
```suggestion
	 * This packet is lower level than the MPEG-TS container, and represents an audio or video
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
+					}],
+				}))
+			}
+		},
+	})
+
+	/**
+	 * Updates ingestion state based on the latest demuxed packet data.
+	 *
+	 * This method is called by our mpegts demuxer, and allows the ingestion engine to track
```
reefdog, line 54:
```suggestion
	 * This method is called by our MPEG-TS demuxer, and allows the ingestion engine to track
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
 			throw new AbstractInstantiationError(this.constructor.name)
 		}
 		this.producer = kafka.producer()
+		this.mpegtsDemuxer = new TSDemuxer(this.processDemuxedPacket)
+		this.mostRecentDemuxedPacket = null
 	}
 
 	/**
-	 * Ingests data into the TV Kitchen pipeline.
+	 * Ingests an mpegts payload into the TV Kitchen pipeline.
```
reefdog, line 30:
```suggestion
	 * Ingests an MPEG-TS payload into the TV Kitchen pipeline.
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
 	 *
-	 * @param  {Buffer} data The buffer of data to be ingested.
-	 * @return {Boolean}     Whether or not the buffer was successfully ingested.
+	 * @param  {Payload} data The payload to be ingested.
+	 * @return {Boolean}      Whether or not the payload was successfully ingested.
```
reefdog, line 33: I think I'm out of my CS depth here, because I am not following how we pass params into `ingestPayloadStream`, and I _think_ this `@return` is now wrong? Isn't it returning a Promise or something? Gah!

Okay so does the `write()` method of the `Writable` object receive the parameters passed to the `ingestPayloadStream` function? If so, I suppose we can rely on people understanding that (to the extent they understand Node streams better than I do), but write now this function is Greek to me and the JSDoc block doesn't seem to align with _reading_ the code, but maybe this does!

So, this is just a request to double-check the comments and the function and give a ✅ that it's in sync as you understand it, since I don't.

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
+	/**
+	 * Process a new chunk of data from an mpegts stream. The chunks passed to this
+	 * function should be presented sequentially, and combine to form a cohearant mpegts
+	 * data stream, but they can be of arbitray size.
 	 *
-	 * @param  {ReadableStream} readableStream The stream of data to be recorded.
-	 * @return {ReadableStream}                The input stream (for chaining)
+	 * This data is used to update local state and ultimately pipe to kafka.
+	 *
+	 * @param  {Buffer} mpegtsData The latest sequential chunk of mpegts data.
```
reefdog, line 79:
```suggestion
	 * @param  {Buffer} mpegtsData The latest sequential chunk of MPEG-TS data.
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
-		.on('data', async (data) => {
-			readableStream.pause()
-			await this.ingestData(data)
-			readableStream.resume()
-		})
+	processMpegtsStream = Transform({
+		objectMode: true,
+		write: (mpegtsData, enc, done) => {
+			this.mpegtsDemuxer.process(mpegtsData)
+			const demuxedPacket = this.getMostRecentDemuxedPacket() || { pts: 0 }
```
reefdog, line 85: Is it worthwhile to move `{ pts: 0 }` to a named constant atop the file like `EMPTY_DEMUXED_PACKET` or maybe `this.generateEmptyDemuxedPacket()`? So we can modify its composition from a more visible place?

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
+			})
+			done(null, payload)
+		},
+	})
 
 	/**
 	 * Start the ingestion engine.
 	 *
-	 * @return {null}
+	 * This creates an ffmpeg process configured to convert input into a predetermined
```
reefdog, line 97:
```suggestion
	 * This creates an FFmpeg process configured to convert input into a predetermined
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
+	getMostRecentDemuxedPacket = () => this.mostRecentDemuxedPacket
+
+	/**
+	 * Process a new chunk of data from an mpegts stream. The chunks passed to this
+	 * function should be presented sequentially, and combine to form a cohearant mpegts
+	 * data stream, but they can be of arbitray size.
 	 *
-	 * @param  {ReadableStream} readableStream The stream of data to be recorded.
-	 * @return {ReadableStream}                The input stream (for chaining)
+	 * This data is used to update local state and ultimately pipe to kafka.
```
reefdog, line 77:
```suggestion
	 * This data is used to update local state and ultimately pipe to Kafka.
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
-		}
-		this.recordStream(ffmpegProcess.stdout)
+
 		logger.info(`Starting ingestion from ${this.constructor.name}...`)
+		inputStream.pipe(ffmpeg.stdin)
+		return pipeline(
+			ffmpeg.stdout,
+			this.processMpegtsStream,
+			this.ingestPayloadStream,
+			() => logger.info(`Ended ingestion from ${this.constructor.name}...`),
```
reefdog, line 116: Would this be the appropriate place to kill the FFmpeg process? (As a separate named function of course.)

src/components/ingestion/__test__/AbstractIngestionEngine.unit.test.js
```diff
@@ -0,0 +1,212 @@
...
+				pts: 1255128,
+				dts: 1252125,
+				frame_ticks: 3003,
+				program: 1,
+				stream_number: 1,
+				stream_id: 224,
+				type: 1,
+				content_type: 2,
+				frame_num: 377,
+			}
```
reefdog, line 161: These sample packets are pretty heavy to have inline, does it violate anything to move them to a `/data/` file? Maybe then import them atop the file (as like `sampleDemuxedPackets`) and here do something like `const demuxedPacket = Object.assign({}, sampleDemuxedPackets[0])`?

src/components/ingestion/AbstractIngestionEngine.js
chriszs: I think this relates to what I was trying to say [here](https://github.com/tvkitchen/tv-kitchen/pull/47#discussion_r432083316).

src/components/ingestion/AbstractIngestionEngine.js
slifty: I saw this as the ingestion engine processing a demuxed packet -- like, not to generate it but more like:

~TSDemuxer Delivery Person Enters Scene~

*Delivery Person (smiling pleasantly):*  Hello there!  Here is `A DEMUXED PACKET`!  Hope you're having a great day.

~Delivery Person proudly places the package on the desk~

*AbstractIngestionEngine Employee (bored):* thanks.  I'll process it now.

~AbstractIngestionEngine Employee takes the packet and proceeds to apathetically slide it into the trash can labeled `MOST RECENT PACKET`~

slifty: Open to swapping `process` with something like `handle` (if maybe `process` evokes the concept of `extract from a deeper stream`)

`ingest` could work as well.

reefdog: Bless you for using narrative dialogue, and explanation format I understand! (That… genuinely did help me to think through it properly.)

I think you can safely leave it `process` given your explanation, though I'd probably err on the side of `handle` or `on` myself. (`complete` only worked in my original one because it was the _packet_ demuxing, not stream demuxing, that was complete… but I agree it is a distractingly final word.)

Anyway, go forth!

src/components/ingestion/AbstractIngestionEngine.js
slifty: Great thought -- I think this.generateEmptyDemuxedPacket would be good... or maybe it would be a good addition to TSDemuxer specifically.

slifty: Update: I can just `import { Packet } from 'ts-demuxer'` and then `new Packet()`

src/components/ingestion/AbstractIngestionEngine.js
slifty: Yes indeed...

And I totally did it as a separate named function *runs back to fix the mistake*

src/components/ingestion/AbstractIngestionEngine.js
reefdog: Hey what about just doing `import { Packet } from 'ts-demuxer'` and then `new Packet()`?

src/components/ingestion/__test__/AbstractIngestionEngine.unit.test.js
slifty: I think that would be very appropriate, yeah

src/components/ingestion/AbstractIngestionEngine.js
slifty: Changing the name to `onDemuxedPacket`

I also added a big docblock for the class itself to describe the overall pipeline (though I do not use narrative dialogue unfortunately)

src/components/ingestion/AbstractIngestionEngine.js
slifty: There is now a `stop()` method which is called at the end of the pipeline (and could be called separately.

src/components/ingestion/AbstractIngestionEngine.js
slifty: Jutin you are right to be confused, this was all wrong.

This is not a method, it is an instance variable.  It shouldn't be `ingestPayloadStream` (which is an imperative) but rather `payloadIngestionStream` (which is a noun).

And yes this pairs directly with Chris's excellent comment!

Thanks all.

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,66 +1,120 @@
...
 	 */
-	recordStream = (readableStream) => readableStream
-		.on('data', async (data) => {
-			readableStream.pause()
-			await this.ingestData(data)
-			readableStream.resume()
-		})
+	processMpegtsStream = Transform({
+		objectMode: true,
+		write: (mpegtsData, enc, done) => {
```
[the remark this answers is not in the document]
slifty: This suggestion was so spot on -- just pushed a rework.

src/components/ingestion/AbstractIngestionEngine.js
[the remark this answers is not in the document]
slifty: I admit I didn't replicate this first, and we really should make integration tests, but I *did* add a `stop` method that gets called at the end of the pipeline so hopefully it's fixed?

slifty:
ALL RIGHT.

I've finally wrangled all the rework from the latest round of comments and I *think* this is close to final form (might even be final form!)

Please give this a re-review at your leisure. Don't worry about giving feedback -- I know this has been a few iterations but this is really important to get right.

I'm also going to be making two issues based on past feedback once we're ready to merge:

1. Moving to `minipass`.
2. Looking into improving the TSDemuxer API so it fits a bit more naturally into our pipeline architecture.

docs/TESTING.md
```diff
@@ -42,6 +42,10 @@ For instance, the following might be a file for an integration test suite of the
 
 Manually created data for unit tests should be stored in a `data` directory within `__test__`.  The same data file should NOT be shared between multiple tests, so as to avoid accidental coupling.
 
+Data should be stored in a json file named `methodBeingTested.json`.  If there is more than one data file for a given tested method it should be appropriately named e.g. `methodBeingTested.moreDetailedDescription.json`.
```
reefdog, line 45:
```suggestion
Data should be stored in a JSON file named `methodBeingTested.json`.  If there is more than one data file for a given tested method it should be appropriately named e.g. `methodBeingTested.moreDetailedDescription.json`.
```

docs/TESTING.md
```diff
@@ -42,6 +42,10 @@ For instance, the following might be a file for an integration test suite of the
...
+
+The data loader only supports json data.
```
reefdog, line 47:
```suggestion
The data loader only supports JSON data.
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+} from 'stream'
 import { Payload } from '@tvkitchen/base-classes'
 import { dataTypes } from '@tvkitchen/base-constants'
 import {
 	AbstractInstantiationError,
 	NotImplementedError,
 } from '@tvkitchen/base-errors'
+import {
+	TSDemuxer,
+} from 'ts-demuxer'
```
reefdog, line 15: Can be collapsed to a single-line import for consistency ([discussed here](https://github.com/tvkitchen/tv-kitchen/pull/52#discussion_r436296822)).

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
 import logger from '%src/lib/logger'
 import kafka from '%src/lib/kafka'
+import {
+	tsToMilliseconds,
+	generateEmptyPacket,
+} from '%src/lib/utils/mpegts'
 
+/**
+ * The AbstractIngestionEngine handles the bulk of the stream processing associated with video
+ * ingestion. Ingeston Engines that implement this class are responsible for defining the input
```
reefdog, line 26:
```suggestion
 * ingestion. Ingestion Engines that implement this class are responsible for defining the input
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+ * stream, and the abstract class will handle coordinating the input pipeline that ultimately
+ * produces messages to the STREAM.CONTAINER kafka queue.
```
reefdog, line 28:
```suggestion
 * produces messages to the STREAM.CONTAINER Kafka queue.
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+ *
+ * The pipeline is consists of the following steps:
```
reefdog, line 30:
```suggestion
 * The pipeline consists of the following steps:
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+ *
+ * 1. The input stream emits video data.
+ * 2. That video data is piped to an FFmpeg spawn process in order to wrap it in an MPEG-TS
+ *    container.
```
reefdog, line 34: This may not matter, but, these steps are not strictly true. The input stream will almost certainly not be emitting raw video data (implied by the language), but some other container format (e.g., MP4). This first FFmpeg process demuxes/remuxes it into MPEG-TS for consistency. It also providing the point at which (if we ever want) we would normalize audio/video codecs.*

I don't think you have to capture all of that, but at least changing "wrap it in an MPEG-TS container" to "ensure the container format is MPEG-TS" (since I would guess that FFmpeg doesn't perform the demux/remux if the container is already MPEG-TS, but that's just a guess).

*ACTUALLY we currently _are_ changing the format by not configuring FFmpeg otherwise. Check this and I'll bring it up elsewhere (since this is important), but by not explicitly passing in the codec pass-through params to FFmpeg, it's almost certainly re-encoding the input to FFmpeg defaults. (Yes, the default FFmpeg behavior in the absence of being told otherwise is to normalize encoding, not pass streams through unmodified.)

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+ * 3. The FFmpeg spawn process emits MPEG-TS data.
+ * 4. That MPEG-TS data is processed by `TSDemuxer` in order to decorate it with the display
+ *    timestamp (DTS).
+ * 5. The decorated data is piped to Kafka as TV Kitchen `Payload`.
+ *
+ * When the stream ends, the FFMPEG process and pipeline is shut down.
```
reefdog, line 40:
```suggestion
 * When the stream ends, the FFmpeg process and pipeline is shut down.
```

You did this one on purpose.

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+ *
+ * The pipeline is consists of the following steps:
+ *
+ * 1. The input stream emits video data.
+ * 2. That video data is piped to an FFmpeg spawn process in order to wrap it in an MPEG-TS
+ *    container.
+ * 3. The FFmpeg spawn process emits MPEG-TS data.
+ * 4. That MPEG-TS data is processed by `TSDemuxer` in order to decorate it with the display
+ *    timestamp (DTS).
+ * 5. The decorated data is piped to Kafka as TV Kitchen `Payload`.
```
reefdog, line 38:
```suggestion
 * 5. The decorated data is piped to Kafka as a TV Kitchen `Payload`.
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+	})
+
+	// A Writeable stream that will ingest Payloads into the TV Kitchen pipeline.
+	payloadIngestionStream = Writable({
+		objectMode: true,
+		write: this.ingestPayload,
+	})
+
+	/**
+	 * Ingests an TV Kitchen payload into the Kafka pipeline.
```
reefdog, line 77:
```suggestion
	 * Ingests a TV Kitchen payload into the Kafka pipeline.
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
 	/**
-	 * Ingests data into the TV Kitchen pipeline.
+	 * Updates ingestion state based on the latest demuxed packet data.
+	 *
+	 * This method is called by our MPEG-TS demuxer, and allows the ingestion engine to track
+	 * the most recent demuxed packet.
 	 *
-	 * @param  {Buffer} data The buffer of data to be ingested.
-	 * @return {Boolean}     Whether or not the buffer was successfully ingested.
+	 * @param  {Packet} packet The latest TSDemuxer Packet object.
```
reefdog, line 104:
```suggestion
	 * @param  {Packet} packet The latest TSDemuxer Packet object
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+		this.mostRecentDemuxedPacket = packet
 	}
 
 	/**
-	 * Continuously ingest of a stream of data.
+	 * Returns the most recent coherent stream packet processed by the ingestion engine.
+	 * This packet is lower level than the MPEG-TS container, and represents an audio or video
+	 * packet demuxed from the MPEG-TS stream.
+	 *
+	 * @return {Packet} The most recent Packet object extracted by TSDemuxer.
```
reefdog, line 115:
```suggestion
	 * @return {Packet} The most recent Packet object extracted by TSDemuxer
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+	 */
+	getMostRecentDemuxedPacket = () => this.mostRecentDemuxedPacket
+
+	/**
+	 * Process a new chunk of data from an MPEG-TS stream. The chunks passed to this
+	 * function should be presented sequentially, and combine to form a coherent MPEG-TS
+	 * data stream, but they can be of arbitrary size.
+	 *
+	 * This method is called by a NodeJS Transform stream and so matches that spec
```
reefdog, line 124:
```suggestion
	 * This method is called by a NodeJS Transform stream and so matches that spec.
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+	 *
+	 * @param  {Buffer} mpegtsData The latest sequential chunk of MPEG-TS data.
```
reefdog, line 126:
```suggestion
	 * @param  {Buffer} mpegtsData The latest sequential chunk of MPEG-TS data
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+	 * @param  {String} enc        The encoding of the passed data.
```
reefdog, line 127:
```suggestion
	 * @param  {String} enc        The encoding of the passed data
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+	 * @param  {Function(err,result)} done     A `(err, result) => ...` callback.
```
reefdog, line 128:
```suggestion
	 * @param  {Function(err,result)} done     A `(err, result) => ...` callback
```

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
-			readableStream.resume()
+	processMpegtsStreamData = (mpegtsData, enc, done) => {
+		this.mpegtsDemuxer.process(mpegtsData)
+		const demuxedPacket = this.getMostRecentDemuxedPacket() || generateEmptyPacket()
+		const payload = new Payload({
+			data: mpegtsData,
+			type: dataTypes.STREAM.CONTAINER,
+			duration: 0,
+			position: tsToMilliseconds(demuxedPacket.pts),
+			createdAt: Date.now(),
```
reefdog, line 139: This emits a UNIX epoch time, which doesn't match our current [Payload definition](https://github.com/tvkitchen/base/blob/master/packages/classes/docs/PAYLOAD.md#attributes), which says it should be ISO 8601. (Smart folks, these `Payload` craftspeople.)

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
 	 */
-	getFfmpegSettings = () => [
-		'-i', this.getInputLocation(),
-		...this.getTranscodeSettings(),
-		'-f', 'mpegts',
-		'-',
-	]
+	stop = () => {
+		if (this.activeIngestionPipeline !== null) {
+			this.activeIngestionPipeline.end()
```
reefdog, line 178: I'm not sure how to test this, but just to sanity-check: this doesn't create a loop, does it? My worry is that by calling `.end()` we may end up triggering this very `stop()` again, and unless something sets `this.activeIngestionPipeline = null` it will keep running. I read a bit about the streams API and couldn't be sure. If you're satisfied it won't (and/or that the tests prove it doesn't), I'll trust you!

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -1,100 +1,209 @@
...
+	 * Returns an FFmpeg settings array for this ingestion engine.
 	 *
-	 * @return {String} The input location to be passed to ffmpeg
+	 * @return {String[]} A list of FFmpeg command line parameters.
 	 */
-	getInputLocation = () => '-'
+	getFfmpegSettings = () => [
+		'-loglevel', 'info',
+		'-i', '-',
+		'-f', 'mpegts',
```
reefdog, line 194: It's nominally outside of the scope of this PR, but it may be extremely worthwhile to go ahead and add the [Stream Copy settings](https://ffmpeg.org/ffmpeg.html#Stream-copy) (`-codec copy`), since currently we are silently changing the encoding of the stream, and that is not intended.

(See earlier comment; default FFmpeg behavior without the stream copy command is to re-encode into default formats, which are something like H264 for video and no idea for audio.)

src/lib/utils/index.js
```diff
@@ -0,0 +1,11 @@
+// Disabling because we intend to have more exports in the future.
+/* eslint-disable import/prefer-default-export */
+
+/**
+ * Checks whether a method exists in a given object
+ *
+ * @param  {String} methodName   The name of the method being checked.
+ * @param  {Object} object       The object that the method should exist in.
+ * @return {Boolean}             Whether the method is defined and is a function.
```
reefdog, line 9:
```suggestion
 * @return {Boolean}             Whether the method is defined and is a function
```

src/lib/utils/index.js
```diff
@@ -0,0 +1,11 @@
+// Disabling because we intend to have more exports in the future.
+/* eslint-disable import/prefer-default-export */
+
+/**
+ * Checks whether a method exists in a given object
```
reefdog, line 5:
```suggestion
 * Checks whether a method exists in a given object.
```

src/lib/utils/index.js
```diff
@@ -0,0 +1,11 @@
...
+ *
+ * @param  {String} methodName   The name of the method being checked.
```
reefdog, line 7:
```suggestion
 * @param  {String} methodName   The name of the method being checked
```

Also, I don't really care, but I don't think we need more than two spaces between a variable name and its description (except when aligning, as below).

src/lib/utils/index.js
```diff
@@ -0,0 +1,11 @@
...
+ * @param  {Object} object       The object that the method should exist in.
```
reefdog, line 8:
```suggestion
 * @param  {Object} object       The object that the method should exist in
```

src/lib/utils/jest.js
```diff
@@ -0,0 +1,13 @@
+// Disabling because we intend to have more exports in the future.
+/* eslint-disable import/prefer-default-export */
+import fs from 'fs'
+import path from 'path'
+
+/**
+ * Loads a file containing json from the test's `data` directory and returns the resulting object.
```
reefdog, line 7:
```suggestion
 * Loads a file containing JSON from the test's `data` directory and returns the resulting object.
```

src/lib/utils/mpegts.js
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable import/prefer-default-export */
+
+/**
+ * Converts a pts or dts extracted from an mpegts packet to seconds.
+ *
+ * @param  {Number} ts       The pts or dts value to be converted.
```
reefdog, line 6:
```suggestion
 * @param  {Number} ts       The pts or dts value to be converted
```

src/lib/utils/mpegts.js
```diff
@@ -0,0 +1,28 @@
...
+ * @param  {Number} baseTime The basetime of the mpegts that the pts came from.
```
reefdog, line 7:
```suggestion
 * @param  {Number} baseTime The basetime of the MPEG-TS that the pts came from
```

src/lib/utils/mpegts.js
```diff
@@ -0,0 +1,28 @@
+/* eslint-disable import/prefer-default-export */
+
+/**
+ * Converts a pts or dts extracted from an mpegts packet to seconds.
```
reefdog, line 4:
```suggestion
 * Converts a pts or dts extracted from an MPEG-TS packet to seconds.
```

src/lib/utils/mpegts.js
```diff
@@ -0,0 +1,28 @@
...
+ * @param  {Number} ts       The pts or dts value to be converted.
+ * @param  {Number} baseTime The basetime of the mpegts that the pts came from.
+ * @return {Number}          The number of seconds represented by the pts or dts.
+ */
+export const tsToMilliseconds = (ts, baseTime = 90000) => +((ts / baseTime) * 1000).toFixed(0)
+
+/**
+ * Exports a TSDemuxer MPEG-TS Packet as defined in the TSDemuxer project.
+ *
+ * @return {Object} A TSDemuxer Packet with no data.
```
reefdog, line 15:
```suggestion
 * @return {Object} A TSDemuxer Packet with no data
```

src/components/ingestion/AbstractIngestionEngine.js
slifty: Ah yeah -- I used to be importing two things; ty!

src/components/ingestion/AbstractIngestionEngine.js
slifty: Good call; will update!

src/components/ingestion/AbstractIngestionEngine.js
slifty: I was probably feeling angry at the time

src/components/ingestion/AbstractIngestionEngine.js
slifty: Good call.  As an aside, I started exploring how to bake type checks into Payload for that kind of thing (this will NOT be the first time we try to assign the wrong thing to createdAt.  In particular I imagine that we'll accidentally assign date objects instead of ISO strings...)

src/components/ingestion/AbstractIngestionEngine.js
slifty: This PR has so much out of scope (e.g. a complete rewrite of the way streams are processed!) that I think this should be fine!

src/lib/utils/index.js
slifty: This ends in a preposition and I'm shocked you didn't catch it.

reefdog: That is one rule up with which I shall not put.

src/components/ingestion/AbstractIngestionEngine.js
slifty: It's a good thought...

But this definitely won't create a callback loop.... because end is not actually a pipeline method, so it will error.  The test didn't catch this because we manually mocked a pipeline and included an `end` method.

I'm going to change this to trigger end on the input stream, which is the correct approach.  I did check and calling `.end()` twice on an input stream does not trigger the pipeline close callback more than once.

reefdog: Wow so I stumbled into helping by shining a light on it for a totally different reason! Like in a heist movie where the security guard drops his flashlight or soda can or something and it rolls behind the very spot the thieves are hiding and he discovers them while retrieving it and then they have to tie him up and escalate the stakes! Like that!!

slifty:
I got the green light from @reefdog to merge this!  GONNA DO IT!

PR #52, opened 2020-05-28
Silences Winston console output during tests

### LINKED ISSUE #49: Winston outputs during tests
#### Description
We probably don't want this kind of output when running tests:

```
 ● Console

    console.log node_modules/winston/lib/winston/transports/console.js:79
      13:15:10 (info): 	Starting ingestion from FullyImplementedIngestionEngine...

    console.log node_modules/winston/lib/winston/transports/console.js:79
      13:15:10 (info): 	Starting ingestion from FullyImplementedIngestionEngine...

    console.log node_modules/winston/lib/winston/transports/console.js:79
      13:15:10 (info): 	Ended ingestion from FullyImplementedIngestionEngine...
```

We're using winston for logging, so we should be able to easily add conditional logging rules depending on whether tests are being run.

#### Related Issues
None

chriszs:
## Description
This PR toggles the silence option on the Winston console transport stream when the `NODE_ENV` is `test`. This environment variable is automatically set when running `yarn test`, so this silences the logger. It's one of several competing solutions discussed in [this StackOverflow question](https://stackoverflow.com/questions/38363292/disable-winston-logging-when-running-unit-tests).

Downsides:

1. it intermingles a test-specific flag with the production code instead of doing so in a Jest setup file;
2. it doesn't follow Jest's `--silent` flag on the command line (should it?).

Other StackOverflow solutions address those downsides. Could investigate adding to a Jest setup further if needed. One line PR was tempting.

## Due Diligence Checklist
- [x] Consolidated commits

## Steps to Test
1. Add `logger.info('whee')` to any test file.
2. Run `yarn test`. You should expect to see no output from the logging statement.
3. Run `yarn test --no-silent`. You should expect to see output from the statement.

## Related Issues
Resolves #49

### DIFF
--- a/docs/TESTING.md
+++ b/docs/TESTING.md
@@ -4,6 +4,17 @@ This project uses Jest for testing.  All new [contributions](docs/CONTRIBUTING.m
 
 **Please review [this excellent guide](https://github.com/goldbergyoni/javascript-testing-best-practices) on JavaScript testing best practices.**
 
+## Test Command
+
+To test the project, run:
+
+```
+yarn test
+```
+
+This will display the test results, but not other logging output from the code by default. To override this add `--no-silent`.
+
+
 ## Test Locations
 
 We have a few types of test, which should be placed in different places depending on the type of test.
--- a/jest.config.js
+++ b/jest.config.js
@@ -2,6 +2,8 @@ const { getModuleAliasesForJest } = require('./lib/utils/moduleAlias')
 
 const settings = {
 	moduleNameMapper: getModuleAliasesForJest(),
+	setupFiles: ['./lib/utils/jest.setup.js'],
+	silent: true,
 }
 
 module.exports = settings
--- a/lib/utils/jest.setup.js
+++ b/lib/utils/jest.setup.js
@@ -1,0 +1,18 @@
+import { NullConsole } from '@jest/console'
+import { transports } from 'winston'
+import logger from '%src/lib/logger'
+
+const getConsoleTransport = () => logger.transports.find(
+	(transport) => transport.constructor.name === transports.Console.name,
+)
+// eslint-disable-next-line no-console
+const isJestSilent = () => console.constructor.name === NullConsole.name
+
+// Respect Jest's silent configuration by silencing Winston's logging.
+// Can be overridden by specifying --no-silent on the command line.
+
+const consoleTransport = getConsoleTransport()
+
+if (consoleTransport) {
+	consoleTransport.silent = isJestSilent()
+}
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
     "winston": "^3.2.1",
     "@tvkitchen/base-constants": "^1.0.0",
     "@tvkitchen/base-classes": "^1.1.0",
-    "@tvkitchen/base-errors": "^1.0.0"
+    "@tvkitchen/base-errors": "^1.0.0",
+    "@jest/console": "^25.1.0"
   },
   "devDependencies": {
     "@babel/core": "^7.8.7",


reefdog:
My picky suggestion for this specific implementation would be to reference the `'test'` environment string through a constants file [like we do here](https://github.com/TechAndCheck/tech-and-check-alerts/blob/3ce0bd99e54ef56323b69e18931f27edf5906e8c/src/server/constants.js#L1-L5), and preferably also to add env-checking utility methods like [here](https://github.com/TechAndCheck/tech-and-check-alerts/blob/3ce0bd99e54ef56323b69e18931f27edf5906e8c/src/server/utils/index.js#L6-L8).

But more substantively, rather than configure silent mode I think what we'd actually want to do is only add the transport when in a non-test environment, sort of like we do with [non-production environments here](https://github.com/TechAndCheck/tech-and-check-alerts/blob/3ce0bd99e54ef56323b69e18931f27edf5906e8c/src/server/utils/logger.js).

Putting this as a comment rather than a review because it's more of a conversation! I agree in principle that we should silence console output during testing.

I played with `jest --silent` briefly by doing a `logger.info` from a test. Without `--silent` the console output happened mid-stream of the test output, while with `--silent` it still output, but after all tests had run. Which is weird and not what we'd want. If it's _possible_ for us to somehow sniff out that `--silent` was passed to Jest and configure Winston accordingly, I _think_ that's a good design? But I don't think it's worth spending too much time on versus just using the test environment as a proxy for this behavior.
Screen Shot 2020-05-29 at 11 07 28 AM

chriszs:
Thanks for the feedback! Super helpful. Will follow it, respond later today or this weekend.

slifty:
One note: when *writing tests* (or debugging a failed test) I sometimes want / need to be able to view console output, so we should make sure whatever solution is landed on accounts for that use case.

chriszs:
Adapted Tech and Check Alert's design, as suggested by @reefdog. But Winston complains when there are no transports. I can add file based logging, then conditionally add console logging _except_ in production and test environments, similar to Tech and Check. But that seems like it's larger than one commit, and possibly this PR.

Questions:
1. Should I do a series of commits?
2. Series of PRs?
3. Is that the behavior we want?

Going to push my WIP commit now so you can see what I mean.

slifty:
So sorry I wasn't able to get to this sooner, but [my power was out for most of this week](https://www.inquirer.com/weather/derecho-storms-philadelphia-deaths-pennsylvania-weather-20200603.html)!

I'll give a review in a moment but as for your questions:

1. I think a series of commits is appropriate; this is more of an art than a science but we've tended to allow our PRs to get a little bigger than ideal at times.

2. That said, if you prefer a series of related PRs that's fine too (just note in the PR that it relies on the merging of a past PR)

3. File logging during tests is where we [wound up in other projects](https://github.com/TechAndCheck/tech-and-check-alerts/issues/83) and I think that's an appropriate way to go here too!

src/lib/logger/index.js
```diff
@@ -4,6 +4,7 @@ import {
 	transports,
 } from 'winston'
 
+import utils from '%src/lib/utils/env'
```
slifty, line 7: In other places we have used named imports for utils -- e.g.

```
import {
  isProductionEnv,
} from '%src/lib/utils/env'
```

Might be good to be consistent there, though also appreciate if maybe that's not good practice somehow.

That said even if you don't do the named import thing, `utils` feels a bit generic as opposed to `env` or `envUtils`.

src/lib/utils/env.js
```diff
@@ -0,0 +1,12 @@
+import config from '%src/config'
+import constants from '%src/constants/env'
+
+const isDevelopmentEnv = () => config.NODE_ENV === constants.ENV_NAMES.DEVELOPMENT
+const isTestEnv = () => config.NODE_ENV === constants.ENV_NAMES.TEST
+const isProductionEnv = () => config.NODE_ENV === constants.ENV_NAMES.PRODUCTION
+
+export default {
```
slifty, line 8: In other util files we use `export const isDevelopmentEnv = ...` etc etc

This pattern eliminates the annoying `prefer export default` problem we face when a utility method only has a single export, but it is a bit more verbose.  That said, it is also a bit more explicit about what is exported.

We should probably pick one and stick with elsewhere in the project.  cc @reefdog -- what do you think?

src/lib/utils/env.js
```diff
@@ -0,0 +1,12 @@
+import config from '%src/config'
+import constants from '%src/constants/env'
```
slifty, line 2: calling this constants feels off to me (similar to calling utils/env `utils` in the earlier method), but I appreciate my gut might be wrong here...

What would you do if you needed to import constants from more than one source?

What if instead you did:

```
import { ENV_NAMES } from `%src/constants/env`
```

src/constants/env.js
```diff
@@ -0,0 +1,9 @@
+const ENV_NAMES = {
```
slifty, line 1: This is not an opinionated question, but I wanted to ask since this is our first constant in the project.

What is the benefit of defining const ENV_NAMES explicitly as opposed to just having something like

```
export default {
  ENV_NAMES: { ... },
}
```

src/lib/logger/index.js
slifty: By the way, I *think* the logic for explicitly pulling the util methods being invoked was it was explicit and as a habit it helps with tree shaking.

src/lib/utils/env.js
slifty: Thinking more about this (and the utils parallel) I do think we should default to pulling the specific items we're using.  Spares us the naming question and also helps us identify dead code some day down the line.

src/lib/logger/index.js
chriszs: Agree! This was how I tried to do it at first and for some reason it wasn't playing nice with the default export I was doing, which in turn was based on a misreading of the `import/prefer-default-export` warning. This is all solved now.

src/lib/utils/env.js
chriszs: Ditto.

src/lib/utils/env.js
chriszs: Changed in latest commit.

src/constants/env.js
chriszs: No difference to me. Could do it this way. Leaving this open for @reefdog's opinion.

slifty: For reef's consideration: we can also do the `export const ENV_NAMES` approach here too.

chriszs: True, and that's my preference, but we'd need something else to export, even if an empty default.

slifty: Ah, for linting right?

Heads up that we regularly use [this pattern]() to handle the linting complaint:

```
// Disabling because we intend to have more exports in the future.
/* eslint-disable import/prefer-default-export */
```

It's a valid linting rule for things like class files, but not so much for utility / constant files that just happen to currently only have one export.

I've been thinking of making it a habit of disabling that rule at the eslintconfig level since 9 out of 10 times I see it complain it's in one of those files.

chriszs: Okay! Thinking about this. Do we intend to have more constants in this file in the future? It's in a file called `env.js`, because it seemed like that fit under a directory called constants. Could either put this in `index.js` or make it the default for this file.

slifty: Ohhhhhhh that's a good point; and yeah that was the right place for it.

And you've convinced me that export default is the way to go here.  Back to deferring to @reefdog on the `const` vs direct export default {} form

reefdog:
My thoughts on "series of commits/PRs": It's totally appropriate to pivot this PR from "silence console logging during tests" to "configure Winston to log properly [in testing]" (with the bracketed bit being optional; even if you tackled proper log configuration for all environments in one PR, that seems like a perfectly contained bit of functionality, even if it involves more LOCs than you estimated at start).

And we're big fans of the laborious but archeologically useful habit of rewriting history in a branch so that we end up with a series of sensibly-contained commits. So in this case, once we like what we're doing here, you _could_ rewind to branch beginning and commit each stage as its own commit, like, "Setup Winston boilerplate", "Configure development logging", "Configure production logging", "Configure test logging", etc. This is just a suggestion, not a requirement (and shouldn't end up being a yak-shaving expedition, am I using that right?) but it is nice to end up with a PR whose component commits allow individual exploration for understanding a larger atomic chunk of functionality being added.

Okay now to actually review!

src/lib/utils/env.js
```diff
@@ -0,0 +1,8 @@
+import config from '%src/config'
+import {
+	ENV_NAMES,
+} from '%src/constants/env'
```
reefdog, line 4: This format is totally fine but I will note that when we're only importing a single item, it's perfectly acceptable to use the one-line `import { ENV_NAMES } from '%src/constants/env'`. We only "enforce" multi-line imports when dealing with 2+ imported items, since it makes diffs so much more legible. (We would then convert to multi-line when importing a second item from this file, which would make a very obvious/legible diff; same when converting from multiple imports to a single one.)

Like I said, this is fine, but you can feel free to collapse it to a single line if you want, too.

src/constants/env.js
reefdog: I'm happy you both defer to me even though I have absolutely no reason to be the authority on this. But now I've done some Googling and while the [ESLint docs themselves were useless](https://github.com/benmosher/eslint-plugin-import/blob/v2.18.2/docs/rules/prefer-default-export.md) as to reasons, there is [this long debate](https://github.com/airbnb/javascript/issues/1365) that was very useful, as well as [this article from the creator of ESLint](https://humanwhocodes.com/blog/2019/01/stop-using-default-exports-javascript-module/). They've convinced me that we should actually purge default exports in favor of named exports; that article, particularly, was convincing.

I think what I'd like to see is:

1. Disable `prefer-default-export`
2. In fact, can we ban them?
3. Convert all our existing default exports to named. The fact that we currently can accidentally `import IApliance from './IAppliance'` is a _bug_ waiting to bite us; it will _work_ in that file (so long as we keep referencing `IApliance`) but things like project-wide searching will break. Instead, by forcing ourselves to use named exports/imports (`import { IAppliance } from './IAppliance`), we avoid this accident. And if for some reason we ever _really_ want to refer to the exported item differently during a specific import, we can always use `import { IAppliance as IApliance } from './IAppliance'`, which is a very explicit way to signal what we're doing.
4. Tangentially, continue pushing off more code organization to file/directory structure, though not so far as the "one export per file".

So for this PR specifically, I'd say remove the `export default {}` bit at the end, add the rule-disabling line that @slifty noted we regularly use, and then let's open a new issue to purge (or at least discuss purging) default exports.

chriszs: Okay. Will do.

chriszs:
Okay, will make this at least two commits.

src/constants/env.js
```diff
@@ -0,0 +1,9 @@
+export const ENV_NAMES = {
+	DEVELOPMENT: 'development',
+	TEST: 'test',
+	PRODUCTION: 'production',
+}
+
+export default {
```
reefdog, line 7: As discussed in the issue (but for some reason not showing in this particular interface, possibly GitHub goofiness) let's purge this default export and add the eslint-disabling boilerplate atop this file.

chriszs:
Note that although this PR adds logging to files, this is generally not my ideological bent, preferring to log to standard and error outputs and let the server environment handle collection and collation of logs, as per [12 factor apps](https://12factor.net/logs):

> A twelve-factor app never concerns itself with routing or storage of its output stream. It should not attempt to write to or manage logfiles. ... In staging or production deploys, each process’ stream will be captured by the execution environment, collated together with all other streams from the app, and routed to one or more final destinations for viewing and long-term archival. These archival destinations are not visible to or configurable by the app, and instead are completely managed by the execution environment.

slifty:
@chriszs regarding that note... you're right X(

Do you think it would be better to:

1. drop ANY type of file logging
2. to have the test command silence output streams of any type and
3. to allow for a 'verbose' test mode instead for cases when you want to see that output

Argh.  I'm sorry we keep running in circles here but I definitely do want to avoid the situation where log files can randomly become 1GB because nobody thought to set up rotation.

chriszs:
Okay, I agree with all that. Here's what I did:

1. I removed the file logging.
2. I went back to silencing the transport in test environments like in the original one line PR (this is necessary because Winston needs at least one transport), but using the new utility function.
3. I added a check so if LOG_LEVEL is set the transport is no longer silenced, allowing you to override to see a specific level of logging.

src/lib/logger/index.js
```diff
@@ -21,6 +22,7 @@ const logger = createLogger({
 	defaultMeta: { service: 'tv-kitchen' },
 	transports: [
 		new transports.Console({
+			silent: isTestEnv() && !config.LOG_LEVEL,
```
slifty, line 25: I'm a bit worried that this is hidden behavior being baked into `LOG_LEVEL` (as opposed to enabling something like `yarn test --verbose`), but I also don't want to belabor things too much and this is a non-critical worry that could easily be addressed if it becomes a priority down the line.

slifty: Note: if we wanted to watch `--verbose` we could do something like

```
silent: isTestEnv() && process.argv.indexOf("--verbose") < 0,
```

slifty: Note: I'm not playing passive disingenuously -- as I think through it I'm leaning to `--verbose` support instead of treating `LOG_LEVEL` as an override (the issue is `LOG_LEVEL` might be set for other reasons than wanting logs in tests.)

But I also appreciate that maybe `--verbose` is non-standard or otherwise bad.

chriszs: Jest has a `--verbose` flag. It "Display[s] individual test results with the test suite hierarchy."

Jest also has a `--silent` flag. It "Prevent[s] tests from printing messages through the console." That's the behavior we want, but it's not the default. We could change `yarn test` to do `jest --silent`, make this behavior parse out the silent flag on the command line (we probably need a new utility for that) and then override by just doing `jest`.

I'd argue using `LOG_LEVEL` makes some sense here. Basically I'm setting a `LOG_LEVEL` of silent (-1?) by default for tests, and then letting you choose a different one when you need it, the same way we have a default of info for other environments and let you override that.

slifty: That makes sense to me!  This also starts to introduce the idea that `LOG_LEVEL` should generally be set on temporary basis for *either* use case.

Is `LOG_LEVEL` documented anywhere?  We should update the documentation to indicate that it will affect test output if set.

Other than that, this is good for merge I think!

reefdog: > Jest also has a `--silent` flag. It "Prevent[s] tests from printing messages through the console." That's the behavior we want, but it's not the default. We could change `yarn test` to do `jest --silent`, make this behavior parse out the silent flag on the command line (we probably need a new utility for that) and then override by just doing `jest`.

I'm not sure if this conflicts with the subsequent `LOG_LEVEL` discussion, but I do think having `jest --silent` disable the console transport is a great idea, and have our `yarn test` call `jest --silent` by default. (As noted we could call `jest` directly to get it non-silent, or we could have a `yarn test:noisy` variant if we wanted. I agree with Chris that we shouldn't overload `--verbose` here.)

reefdog: tbh I don't follow the `LOG_LEVEL` suggestions; I don't even know _how_ to set environment variables for the test environment, except as inline parameters in the yarn script definition.

reefdog:
(Separate from review, wanted to note that I did not know about 12-factor apps, and really like [Chris's guidance](https://github.com/tvkitchen/tv-kitchen/pull/52#issuecomment-640124385) to not define file logging within the application code. It's deeply convenient but also yes can introduce quicksand.)

src/lib/logger/index.js
slifty: Following the `--silent` thread -- the reminder that we can, in fact, just call yarn jest in situations where we want to skip `--silent` is a really good point that I had not been considering, because I am a fool..

(side note we could also have a `yarn test:debug` if we really wanted to have that point documented, but meh).

chriszs: Doing some research on the behavior of silent. I'm curious about how it's supposed to work and why it doesn't with Winston. It does _something_ because as @slifty mentioned there's a difference between how the log output appears when you add the flag versus when you don't, but unlike `console.log` it doesn't succeed in suppressing it. More important, there's a `--no-silent` flag you can append after `--silent` to counteract it, like some sort of counter-spell that just solves the whole problem for us.

chriszs: Okay, so this is actually a weird interaction between Jest and Winston behavior:

Act 1: Jest normally hijacks the `console` with its own version called `BufferedConsole`, which pretends to be the `console` we know and love by providing its own writable stream instead of the standard output stream.

Act 2: Meanwhile, Winston's Console transport writes to `console._stdout` if it exists, `console.log` if it doesn't.

All good so far.

Act 3: But if we pass the `--silent` flag Jest instead subs in `NullConsole`, which is just a bunch of no-ops where logging functions should go. `stdout` doesn't get hijacked, so it just outputs.

* audience gasps *

Okay, so, what does this mean? It's possible we have the basis for Jest PR. But in the meantime, we could detect the `silent` flag by testing whether `console` is an instance of `NullConsole`. This is really testing specific, though, so it could go in a setup file.

reefdog: Excellent detective work, but just to check: we can't suss out `jest --silent` with `process.argv.indexOf("--silent") < 0` in a utility method? (I don't know if `process.argv` stays populated all the way through the… call stack? Am I using that word right??)

chriszs: I was going to reply with "yes of course we can but where's the fun in that," except, it appears no! We can't, because Jest runs the tests in a worker process that doesn't get the original args. Here's what I get in `argv` anyway:
```
/Users/chriszs/.nvm/versions/node/v14.4.0/bin/node,/Users/chriszs/Desktop/work/tv-kitchen/node_modules/jest-worker/build/workers/processChild.js
```
Thank god all that sleuthing had a purpose.

chriszs: Okay, just did as I suggested. See what you think. Assume we're not going to require JSDoc, unit tests for a test setup file, but can add those if needed.

slifty:
Wow -- chris this is a *much* better solution to what's out there on the internet to date for solving this problem.  Amazing stuff... and boy would it be cool if this did lead to a PR to Jest.

At the very least I'd urge you to weigh in on some of the StackOverflow questions related to this!

I agree that this does not need unit testing.

I would throw in one little note in the TEST.md documentation to document that jest tests in silent mode, and that folks who need console out should run `yarn test --nosilent` (assuming that is the command)

No need for re-review.

chriszs:
Oh, there's a TEST.md. So that's where I was supposed to document this. Will do.

docs/TESTING.md
```diff
@@ -4,6 +4,17 @@ This project uses Jest for testing.  All new [contributions](docs/CONTRIBUTING.m
...
 
+## Test Command
+
+To test the project, run:
+
+```
+yarn test
+```
+
+This will display the test results, but not other logging output from the code by default, to override this add `--no-silent`.
```
reefdog, line 15:
```suggestion
This will display the test results, but not other logging output from the code by default. To override this add `--no-silent`.
```

chriszs: Done.

PR #56, opened 2020-06-07
Create a file ingestion engine

### LINKED ISSUE #21: Create a file ingestion engine
## Task
The ability to send a video file into the TV kitchen is intended to be more of a developer tool for now, though eventually it may obviously have value to end users.

Given that, it is OK for the file ingestion engine to only work with a predefined file type (e.g. MP4).

#### Description
Create the file ingestion engine along with relevant tests.

No.

#### Related Issues
Issue #15

chriszs:
## Description
This creates a FileIngestionEngine. It is a concrete implementation of AbstractIngestionEngine, which reads in a file, converts it into an MPEG-TS stream represented as a series of Payloads, and then pipes that to Kafka. This PR mostly defines the input stream as a file stream.

It also adds an integration test for file ingestion and fixes some errors in the AbstractIngestionEngine uncovered by this test.

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits

## Steps to Test
1. `yarn test`

## Related Issues
Resolves #21, #57, #58, #59, #60, #61


src/tests/ingestion.int.test.js
```diff
@@ -0,0 +1,39 @@
...
+// Test imports
+import path from 'path'
+import { PassThrough } from 'stream'
+import FileIngestionEngine from '%src/components/ingestion/FileIngestionEngine'
+
+jest.mock('child_process')
+
+describe('ingestion', () => {
+	describe('FileIngestionEngine', () => {
+		it('it should read a video file into a Kafka topic', () => {
```
slifty, line 13: My only thought is "should this integration test mock kafka / should we actually test that kafka is working?"

But I also think that could be a separate test that we write some day and we don't need to add that complexity to this PR.

slifty: For instance should we at some point have an integration test that:

1. Ingests the file into kafka.
2. Reconstructs the file from kafka.
3. Ensures they're the same file.

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -139,6 +114,33 @@ class AbstractIngestionEngine {
 		done(null, payload)
 	}
 
+	// The FFmpeg process used to wrap the ingestion stream in an MPEG-TS container
+	ffmpegProcess = null
```
slifty, line 118: Very happy this all works now, but something feels wrong about having all of this stuff below instance methods.

For context: I had moved things out of the constructor for the sake of readability, but having things under methods makes me concerned they're a bit buried.

This is a genuine question: might it be better to have the nulled versions of these instances defined above the constructor, and then populate them in the constructor?

Does that even solve the problem, or do we face the same order challenges?

TL;DR -- I'm fine with leaving this as is and don't want to spend a million years moving it all around forever and ever, just wanted to raise it in case someone had a Strong Opinion

src/tests/ingestion.int.test.js
chriszs: Good question! I wrestled with whether to mock that and ffmpeg. I think a test that does what you describe could be valuable, I also wonder whether the overhead of that will result in slower, brittler tests. But having coverage for the Kafka integration could be valuable. I think an interim solution is to add a little more specificity to what we’re expecting send to be called with, assuming send is going to work.

slifty: Good call -- and maybe we'd want to get better with test tags before adding a test like that (so we could have `yarn test` default to just the fast ones and then like `yarn test:i:have:lots:of:time` for doing the true full suite...

src/components/ingestion/AbstractIngestionEngine.js
chriszs: Moved nulled properties above the constructor, set them in the constructor.

src/tests/ingestion.int.test.js
chriszs: Made this test compare snapshots of the payloads passed to `send`.

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -38,38 +38,46 @@ import {
...
+	mpegtsProcessingStream = null
 
 	// A Writeable stream that will ingest Payloads into the TV Kitchen pipeline.
-	payloadIngestionStream = Writable({
-		objectMode: true,
-		write: this.ingestPayload,
-	})
+	payloadIngestionStream = null
+
+	constructor() {
```
slifty, line 62: Excellent -- this approach feels just right thank you for iterating!

src/components/ingestion/FileIngestionEngine.js
```diff
@@ -0,0 +1,33 @@
+import fs from 'fs'
+import AbstractIngestionEngine from '%src/components/ingestion/AbstractIngestionEngine'
+
+/**
+ * The FileIngestionEngine handles processing a video file. It is a concrete
+ * implementation of AbstractIngestionEngine, which reads in a file,
+ * converts it into an MPEG-TS stream represented as a series of Payloads,
+ * and then pipes that to Kafka.
+ *
+ * @param  {string}  path The path of the file to be ingested
```
slifty, line 10: I don't know how JSDocs actually work so take this with a grain of salt, but [this example](https://jsdoc.app/howto-es2015-classes.html) seems to have constructor documentation as part of the constructor directly.

It also includes an `@extends` keyword we might want to consider here.

src/components/ingestion/FileIngestionEngine.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+class FileIngestionEngine extends AbstractIngestionEngine {
+	constructor(path) {
+		if (!path) {
+			throw new Error('path required') // TODO: should probably be FileInstantiationError
```
slifty, line 15: Just noting since there's a TODO here -- should it be or not?  If it should, should we just do that here?  If not, should we take out the TODO?

From what I can tell, the benefit of custom errors is:

1. It lets you specify extra metadata (probably not relevant in this case)
2. It allows for more specific try / catch blocks in general.

It might also be good fodder for `CONTRIBUTING.md` if we land on a project philosophy around errors.

Note that we *would not have to add this error to base-errors*.  That package is for errors that are thrown by `base-classes` or `base-interfaces`.  The point being: we could have a custom `error` in this repository directly (which is just a loser barrier to entry.

I'm comfortable moving in that direction now -- including an update to CONTRIBUTING.md in this PR with a note akin to "Never use a vanilla `Error`, always try to use an existing custom error type.")  Unless we hate that!  In which case we should probably just document what we want.

I do think that the custom error should probably be a slightly more generic `InstantiationError` rather than `FileIngestionInstantiationError` (unless `FileInstantiationError` is a thing and you meant it!).

src/tests/ingestion.int.test.js
```diff
@@ -0,0 +1,45 @@
...
+
+describe('ingestion', () => {
+	describe('FileIngestionEngine', () => {
+		it('it should read a video file into a Kafka topic', (done) => {
+			jest.clearAllMocks()
+
+			const mockDate = new Date(1592156234000)
+			jest
+				.spyOn(global, 'Date')
+				.mockImplementation(() => mockDate)
```
slifty, line 19: oh snap this is great ty for showing this approach

src/tests/ingestion.int.test.js
```diff
@@ -0,0 +1,45 @@
...
+			ingestionEngine.producer = {
+				send: jest.fn().mockResolvedValue(),
+				connect: jest.fn().mockResolvedValue(),
+				disconnect: jest.fn(),
+			}
+
+			ingestionEngine.start()
+
+			ingestionEngine.producer.disconnect.mockImplementation(() => {
+				expect(ingestionEngine.producer.send).toMatchSnapshot()
```
slifty, line 40: Ty for adding this to the test!

Would it be too far to *additionally* (or maybe alternatively, depending on what you think after reading this) re-combine all of the buffers of the resulting calls, and compare that buffer to the original video?  Basically, we want to make sure that the original video was sent in order and that the resulting payload stream reconstructs that video.

I am worried that it will break regularly as we change what Payloads look like.

That's fine (re-snapshotting is just a thing we will be doing), but I am also worried that the circumstances that re-snapshotting happen are also the ones where we are relying on tests to ensure nothing broke (e.g. the re-snapshot will pass, but the new snapshot might be broken without us realizing).

src/tests/ingestion.int.test.js
chriszs: Thank you! I got it from StackOverflow.

src/tests/ingestion.int.test.js
chriszs: Ah, I see what you're saying. That feels like a lot of logic for a test, and snapshots seem like a really clean way to deal with this, but maybe there's a short cut way to do that. Will think on it.

src/components/ingestion/FileIngestionEngine.js
chriszs: I was looking to the errors in AsbtractIngestionEngine for guidance. Happy to do whatever, but agree adding it to the base package and defining a new error for every circumstance (as opposed to something slightly more generic) seemed excessive, which was why it was a TODO.

src/components/ingestion/FileIngestionEngine.js
chriszs: Okay. Will change.

src/components/ingestion/FileIngestionEngine.js
slifty: I was googling "does anybody care about custom errors in Node" and got basically zero results so I feel lost and afraid.

src/tests/ingestion.int.test.js
reefdog: feature request that `yarn test:i:have:lots:of:time` loads up CATS (2019) dir. Tom Hooper in the background and takes the entire runtime to complete.

src/components/ingestion/AbstractIngestionEngine.js
```diff
@@ -84,12 +92,14 @@ class AbstractIngestionEngine {
...
 		} else {
-			done(this.producer.send({
+			this.producer.send({
 				topic: dataTypes.STREAM.CONTAINER,
 				messages: [{
 					value: Payload.serialize(payload),
 				}],
-			}))
+			})
+				.then((result) => done(null, result))
```
reefdog, line 101: Super tiny thing, but I generally like to keep the first method on the same line as the Promise's closing paren, so this becomes
```
}).then()
  .catch()
```
But I'm not terribly wedded to that. Is this a thing we should care about, let be totally up to personal preference, or make circumstantial?

src/components/ingestion/FileIngestionEngine.js
```diff
@@ -0,0 +1,33 @@
+import fs from 'fs'
+import AbstractIngestionEngine from '%src/components/ingestion/AbstractIngestionEngine'
+
+/**
+ * The FileIngestionEngine handles processing a video file. It is a concrete
+ * implementation of AbstractIngestionEngine, which reads in a file,
+ * converts it into an MPEG-TS stream represented as a series of Payloads,
```
reefdog, line 7: Chris I can't tell you what a relief it is to read "an MPEG-TS stream" without having to go through three review cycles. VERY MUCH CC @slifty

src/components/ingestion/FileIngestionEngine.js
reefdog, line 10: I think Dan mentions this already (it's annoying that I'm not seeing his review comments in this tab), but yeah I thiiiink we JSDoc the constructor and not the top of the class? (Update: yes!)

src/components/ingestion/FileIngestionEngine.js
reefdog, line 15: I think Dan covers this already (let's either use the custom error type or remove the TODO, preference for the former), but either way the error message should be more descriptive.

src/components/ingestion/FileIngestionEngine.js
```diff
@@ -0,0 +1,33 @@
...
+		if (!path) {
+			throw new Error('path required') // TODO: should probably be FileInstantiationError
+		}
+		super()
+
+		this.path = path
+	}
+
+	// The path of the file to read by the engine
+	path = null
```
reefdog, line 23: This should be moved above the constructor for the reasons decided in `AbstractIngestionEngine`, right?

src/components/ingestion/AbstractIngestionEngine.js
chriszs: Will do.

reefdog: Oh wait I'll bet this is because alignment of the `then/catch` methods is hard because we're using tabs instead of spaces and thus can't guarantee where they align. If that's the case and what's driving this then I retract my request!

src/components/ingestion/FileIngestionEngine.js
chriszs: You’ll love my secret refactor then.

src/components/ingestion/FileIngestionEngine.js
chriszs: Okay!

src/components/ingestion/FileIngestionEngine.js
reefdog: :ohno:

src/components/ingestion/FileIngestionEngine.js
chriszs: “Path absolutely required?”

src/components/ingestion/FileIngestionEngine.js
chriszs: Uh sure.

src/components/ingestion/FileIngestionEngine.js
reefdog: I was thinking more like "FileIngestionEngine must be instantiated with a file path"

chriszs: Done.

src/components/ingestion/FileIngestionEngine.js
chriszs: Done.

src/components/ingestion/FileIngestionEngine.js
chriszs: Done.

src/components/ingestion/AbstractIngestionEngine.js
chriszs: Ran in through prettier. The result both lints and looks great. Don't tell Dan.

src/tests/ingestion.int.test.js
chriszs: Did this.

src/components/ingestion/FileIngestionEngine.js
chriszs: Based on the conversation in Slack, I did not define a custom error. Should I have?

src/tests/ingestion.int.test.js
chriszs: Denied.

src/components/ingestion/FileIngestionEngine.js
slifty: That was the right call!

src/components/ingestion/AbstractIngestionEngine.js
slifty: PRETTIERS?!?!?!?!?!?!

src/components/ingestion/FileIngestionEngine.js
slifty: (Is there a linting rule we could enable for this 🤔 )

src/components/ingestion/FileIngestionEngine.js
slifty: Clearly we just pronounce MPEG very differently justin.

reefdog: I sing it like Hanson.

Issue #54, opened 2020-06-06
Make Docker image, docker compose for project


chriszs:
#### Describe the solution you'd like
It'd be good if this project had an official Docker image and Docker Compose file for running the project. Dockerization may be a good approach for TV Kitchen, because people will want to run this in varied environments and Docker allows them to do that without additional setup or modification to those environments, making the project easier to run and manage. In the future, these may include Kubernetes or the various AWS services that support containers.

My suggested plan of action would be:

- [ ] Add a Dockerfile
- [ ] Add a Docker build step in Travis
- [ ] Move docker-compose file to project's top level (this is convention), add TV kitchen image to it
- [ ] Switch Kafka start command to do `docker-compose up -d kafka`
- [ ] Document Docker compose option in README
- [ ] Use Travis to push an official image to a TV Kitchen Docker Hub project on commits, version releases

Can do a quick PR with this if we think this is a good plan.

chriszs:
Reframed this to a feature request, though unable to the modify the label.

reefdog:
I'm all for it! Would hitting #29 be a natural fit at the same time?

chriszs:
Looks like it. Docker Compose [has ways to use environment variables](https://docs.docker.com/compose/environment-variables/) and should pull from an `.env` file in the working directory.

slifty:
This issue was created before we moved to the library architecture; now that we have that, the concept of docker containers are more relevant to the recipes and appliances directly.

We may still find that the countertop should have its own docker container, but for now I think we can close this?  (Please feel free to re-open if I'm totally wrong!)

Issue #50, opened 2020-05-28
Move from Node 13.x


slifty:
#### Description
@chriszs just pointed out that Node 13 is reaching EOL [in a few days](https://nodejs.org/en/about/releases/) and that Node 14 will be the next LTS.

We should either move to Node 14 or back to Node 12.

#### Relevant Resources / Research
No

#### Related Issues
No

PR #63, opened 2020-06-13
Add Dockerfile

chriszs:
This adds a Dockerfile and .dockerignore file to allow the project to be built as a docker image. This is the first of several incremental changes for Docker readiness outlined in #54.

## Due Diligence Checklist
- [x] Consolidated commits

## Steps to Test
1. `docker build .`
2. `docker run --env-file=.env <image hash>`

## Related Issues
Begins work outlined in #54

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,14 @@
+FROM node:14-alpine
+
+WORKDIR /usr/src/app
+
+COPY package.json ./
+COPY yarn.lock ./
+
+RUN yarn
+
+COPY . .
+
+RUN apk add  --no-cache ffmpeg
+
+CMD [ "npm", "start" ]
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+node_modules
+.env
+logs
+*.log


Dockerfile
```diff
@@ -0,0 +1,14 @@
...
+COPY package.json ./
+COPY yarn.lock ./
+
+RUN yarn
+
+COPY . .
+
+RUN apk add  --no-cache ffmpeg
+
+CMD [ "npm", "start" ]
```
slifty, line 14: Should we have it run kafka too? or were you thinking kafka would be loaded in a separate container and docker-composed together?

chriszs: Oh, separate container! Docker Composed together.

chriszs: Modifying Docker Compose is step 3 in my multi-step plan.

Dockerfile
```diff
@@ -0,0 +1,14 @@
+FROM node:14-alpine
```
reefdog, line 1: The answer to this question is probably tied up in #50, but is it okay if this drifts from the version specified in our `package.json` (currently 13)? I have no particular Node version allegiance, but a slight preference for LTS > current (unless we have specific features we need that aren't in LTS). Mostly just curious if we need to ensure this is in sync with the package file.

chriszs: `package.json` specifies a minimum, so I think exceeding it is okay. But Travis is also set to test on 13, so there are a couple signals 13 is the current default. It's also EOL now, which we mostly shouldn't care about, except that we want the Docker image and everything else to have the longest functional life possible.

slifty:
Sanity check: given some of the discussion in #66 should we hold off on merging this? (if not, I think we're probably OK to merge)

chriszs:
Closing for now. I think we will want a Dockerfile somewhere at some point, so I'm leaving the branch.

PR #75, opened 2020-07-04
Create HTTP ingestion engine

### LINKED ISSUE #73: Create HTTP ingestion engine
#### Description
We need an HDHomeRun ingestion engine, which streams MPEG-TS over HTTP. Therefore, it should make sense to write a generic HTTP ingestion engine first that the HDHomeRun ingestion engine can inherit from and decorate (see [this comment](https://github.com/tvkitchen/tv-kitchen/issues/68#issuecomment-650600592)). This way we write the most minimal engine possible, test HDHomeRun functionality before enhancing it with HDHomeRun-specifics and accommodate  other HTTP streaming sources (really any TS file on a server).

- [HDHomeRun HTTP Development Guide](https://www.silicondust.com/hdhomerun/hdhomerun_http_development.pdf)

#### Related Issues
#68

chriszs:
## Description
Creates a HTTP ingestion engine that can serve as a base for a HDHomeRun engine.

## Due Diligence Checklist
- [x] Tests
- [x] Consolidated commits

## Steps to Test
1. `yarn test`

## Related Issues
Resolves #73

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,8 @@
     "@tvkitchen/base-constants": "^1.0.0",
     "@tvkitchen/base-classes": "^1.1.0",
     "@tvkitchen/base-errors": "^1.0.0",
-    "@jest/console": "^25.1.0"
+    "@jest/console": "^25.1.0",
+    "node-fetch": "^2.6.0"
   },
   "devDependencies": {
     "@babel/core": "^7.8.7",
--- a/src/components/ingestion/HttpIngestionEngine.js
+++ b/src/components/ingestion/HttpIngestionEngine.js
@@ -1,0 +1,51 @@
+import { PassThrough } from 'stream'
+import fetch from 'node-fetch'
+import AbstractIngestionEngine from '%src/components/ingestion/AbstractIngestionEngine'
+
+/**
+ * The HttpIngestionEngine handles processing a HTTP video stream. It is a concrete
+ * implementation of AbstractIngestionEngine, which reads in a stream,
+ * converts it into an MPEG-TS stream represented as a series of Payloads,
+ * and then pipes that to Kafka.
+ *
+ * @extends AbstractIngestionEngine
+ */
+class HttpIngestionEngine extends AbstractIngestionEngine {
+	// The URL of the stream to read by the engine
+	url = null
+
+	/**
+	 * Create a HttpIngestionEngine.
+	 *
+	 * @param  {string}  url The url of the stream to be ingested
+	 */
+	constructor(url) {
+		super()
+		if (!url) {
+			throw new Error(
+				'HttpIngestionEngine must be instantiated with a stream URL',
+			)
+		}
+
+		this.url = url
+	}
+
+	/**
+	 * @inheritdoc
+	 */
+	getInputStream = () => {
+		const stream = new PassThrough()
+
+		fetch(this.url)
+			.then((response) => {
+				response.body.pipe(stream)
+			})
+			.catch((err) => {
+				stream.emit('error', err)
+			})
+
+		return stream
+	}
+}
+
+export default HttpIngestionEngine
--- a/src/components/ingestion/__test__/HttpIngestionEngine.unit.test.js
+++ b/src/components/ingestion/__test__/HttpIngestionEngine.unit.test.js
@@ -1,0 +1,48 @@
+import nock from 'nock'
+import HttpIngestionEngine from '../HttpIngestionEngine'
+
+describe('HttpIngestionEngine', () => {
+	describe('constructor', () => {
+		it('should throw an error when called without a URL', () => {
+			expect(() => {
+				new HttpIngestionEngine() // eslint-disable-line no-new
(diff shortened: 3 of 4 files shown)


package.json
```diff
@@ -42,7 +42,8 @@
     "@tvkitchen/base-constants": "^1.0.0",
     "@tvkitchen/base-classes": "^1.1.0",
     "@tvkitchen/base-errors": "^1.0.0",
-    "@jest/console": "^25.1.0"
+    "@jest/console": "^25.1.0",
+    "node-fetch": "^2.6.0"
```
slifty, line 46: We've been tending to put new dependencies in their own commit (e.g. "Add node-fetch package") and I think that would be good to continue in this PR.

I appreciate that you explain a bit in this commit as to why we're using node-fetch vs something native in node but a quick extra note would be helpful (like, one sentence; not re-producing an entire "fetch vs http" blog post or whatnot)!

src/components/ingestion/HttpIngestionEngine.js
```diff
@@ -0,0 +1,53 @@
...
+	 *
+	 * @param  {string}  url The url of the stream to be ingested
+	 */
+	constructor(url) {
+		if (!url) {
+			throw new Error(
+				'HttpIngestionEngine must be instantiated with a stream URL',
+			)
+		}
+		super()
```
slifty, line 28: In Java I believe you *must* call super first; I appreciate that we don't need to reference `this` in our validation here, and so validation first works, but in the general case it's *possible* a validation rule might rely on the property of an inherited class and for that reason maybe standardizing around calling `super()` right away is worthwhile.

(if you disagree, plz disregard this comment!)

package.json
slifty: I know tweaking history is not a common pattern so here's the steps in case they're handy:

* `git rebase -i`
* mark this commit as "edit"
* create a *NEW* commit (while rebasing) that contains just the package + yarn.lock
* continue the rebase with the rest of the commits

src/components/ingestion/HttpIngestionEngine.js
```diff
@@ -0,0 +1,53 @@
...
+
+		this.url = url
+	}
+
+	/**
```
slifty, line 33: Since getInputStream is an override method, rather than replicate the documentation lets use the [`@inheritdoc`](https://jsdoc.app/tags-inheritdoc.html) tag.

NOTE: I think I need to make an issue about this since we're totally not doing that elsewhere.

src/components/ingestion/__test__/HttpIngestionEngine.unit.test.js
```diff
@@ -0,0 +1,49 @@
...
+			})
+
+			const fileEngine = new HttpIngestionEngine(
+				url,
+			)
+
+			const inputStream = fileEngine.getInputStream()
+
+			inputStream.on('error', (err) => {
+				expect(err.message).toEqual(`request to ${url} failed, reason: ${message}`)
```
slifty, line 44: I know the drama that led to checking messages here, but I have concern that the content of an error message is too rigid for a test (e.g. the spec is not that the message looks like this exactly, but rather that the error is of the expected type)

src/components/ingestion/__test__/HttpIngestionEngine.unit.test.js
```diff
@@ -0,0 +1,49 @@
...
+				done()
```
slifty, line 45: If no error is sent this test will fail but it will fail based on timeout settings, right?  I would prefer a more explicit test that will fail directly if no error has been thrown (who knows if we will have to play with test timeout settings in some environment some day).

Would it be reasonable to modify this test so the `error` event sets a value and the `finish` event is the place that actually runs the `expect` statements in terms of those values?  For instance, having `error` append the error to an `errorCache` array, and then in `finish` have something like:

`expect(errorCache.length).toBe(1)`

chriszs: I looked into this, but problem I believe I was having that it didn't emit finish. My research at the time indicated that the error event ended the stream, emitting nothing else, and this was expected and normal. But I'll check again.

chriszs: Okay, used Jest's promise resolution/rejection support for this. Seems better.

src/components/ingestion/__test__/HttpIngestionEngine.unit.test.js
chriszs: Changed.

src/components/ingestion/HttpIngestionEngine.js
chriszs: Okay.

src/components/ingestion/HttpIngestionEngine.js
chriszs: Changed.

package.json
chriszs: Okay.

Issue #80, opened 2020-07-14
Pick a package name and rename this repo


slifty:
#### Description
As we have fleshed out this project we have created a few repositories and started to shift to treating this one as more of an API than a standalone application (#66).

When we finally finish that API and publish a package, we will want to consider renaming this repository to match the package name.

To be clear, #66 should be finished first, but I wanted to open this issue so we don't lose track of the renaming task.

#### Relevant Resources / Research
None.

#### Related Issues
Issue #66

slifty:
Some brainstormed ideas for the ultimate package name (and therefore repo name)

- @tvkitchen/kitchen
- @tvkitchen/core (which will be confused with base)
- @tvkitchen/countertop (which is great, but technically it includes ingestion engines which are CURRENTLY separate from the countertop but there is talk about [whether ingestion engines are appliances](https://github.com/tvkitchen/meta/issues/14))
- ...

slifty:
Now that we're moving ingestion engines to be appliances, I think It is time to call the `@tvkitchen/countertop`.

We can also remove the `components` directory, since the countertop *is* the package.

slifty:
It is done!  `@tvkitchen/countertop`

Issue #66, opened 2020-06-15
What should the TV Kitchen entry point be?


slifty:
#### What do you want to talk about?

First: @chriszs has started working on a Docker entry point (🎉) over at #54, and we should continue that line of exploration to think a bit about how people are going to run a copy of TV Kitchen.

We should think about what we want to do for the short term, what we want to support for the long term, and what we want to document as the "recommended" way.  Note that we can support more than one deployment technique.

Some options as I see them right now:

1. `git clone` of the `tv-kitchen` repository and following configuration steps.
2. Docker container (e.g. #54) along with configuration steps.
3. A `tvk` command line tool (e.g. `tvk start -c config.json` or whatever).
4. An npm package that people can import into their own code base.

Some of this came up in dialogue around #65, since some of the appliance installation approaches might have relied on a modification of a `package.json` (we are likely NOT going to be following those approaches).

#### Relevant Resources / Research
None for now, though I know there must be a ton out there about this issue.

chriszs:
A couple projects I've done have followed a pattern of script (#1) -> API (#4) -> CLI (#3) -> Docker (#2), where the script is an intermediate stage, later refactored to a CLI sitting on top of an API, installable via `npm` or `npx` and then Dockerized.

chriszs:
An additional consideration: Okay, so you have your daemon. Let's say it has a basic CLI focused on providing file-based configuration. Do you also want a separate configuration CLI that acts as a client once you have a configuration store? Or is that the same CLI?

reefdog:
> 3. A `tvk` command line tool (e.g. `tvk start -c config.json` or whatever).

So presumably what this might do is create and fill a local instance from a template. (Also presumably another way to get that template would be to just clone it directly from GitHub rather than use the CLI tool.) That template would have the main tv-kitchen repo (this one, call it TVK core) as a dependency.

That's kind of the model I was imagining, personally.

chriszs:
Oh, interesting! So like Yeoman?

reefdog:
_Yeoman. Now there's a name I've not heard for a long time. A long time._

But, yes, kind of? Basically the architecture I'm assuming is that the unique footprint of an individual TV Kitchen installation should be as small as possible, mostly config files, with TV Kitchen core, appliances, etc being imported as dependencies.

[...]

PR #88, opened 2020-08-26
Reorganize as countertop package

### LINKED ISSUE #85: Reorgnaize repository to be a package
#### Description
We're moving to releasing this as a package but to do that we need to improve the repository a bit:

1. Remove the concept of `components` (this is going to ONLY be the countertop)

We may also want to remove the use of environment variables for ANY form of configuration, but I will treat that as a separate issue.

None

#### Related Issues
Issue #80

slifty:
## Description
This PR is a significant change: it moves this repository from being a standalone "application" that can be cloned, configured, and run into being a "countertop" package that a TVK implementation will import (alongside the other appliances / components that it wants to use).

This is the culmination of a series of discussions about what the TVK entry point should look like.

It means we no longer have a `src/components` directory, since this is *only* the countertop.  The countertop classes are now in a `src/classes` directory instead.

We expose all of the classes at the package level, so developers will do things like:

```
import { CountertopCoordinator } from `@tvkitchen/countertop`
const countertop = new CountertopCoordinator()
```

Questions for the room:

1. Should `CountertopCoordinator` just be called `Countertop` (very important)
2. Is `src/classes` the right directory (less important) 
3. Is `@tvkitchen/countertop` an appropriate package name (last call to change it!).

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits
- [x] Changelog(s) updated

## Steps to Test
1. `yarn test`

## Deploy Notes
None -- deployment is no longer really a thing here.

## Related Issues
Resolves #85 
Related to #66 
Related to #80


slifty:
NOTE: part of merging this PR is going to be renaming the repository to `countertop`

reefdog:
> Should `CountertopCoordinator` just be called `Countertop` (very important)

**Yes!** I think `CountertopCoordinator` is like saying "we wanted to go with the metaphor, but found the name illegible, so what it really is is a coordinator, so we'll call it both". I'd say call it _either_ the `Countertop` _or_ the `ApplianceCoordinator` (or whatever _thing_/action it is coordinating). IIRC it's doing enough general coordinating that adding a subject could go out of date quickly, so, `Countertop` works as a general name that's clear _enough_ and whose legibility is quickly established once you grok the architecture.

> Is src/classes the right directory (less important)

**I have no opinion!** Semantically it makes sense, but I don't know if there's an idiomatic reason to go with another name!

> Is @tvkitchen/countertop an appropriate package name (last call to change it!).

**Yes!**

README.md
```diff
@@ -1,29 +1,31 @@
...
+# TV Kitchen Countertop
 
 The [TV Kitchen](https://tv.kitchen) is a playground that turns TV into data (transcripts, facial recognition, etc.) so that newsrooms can help their staff and readers better understand what is happening across various local and cable TV news outlets.
 
+The Countertop is the entry point for developers who want to set up a TV Kitchen.
+
 ## Setting up the project
 
-If you are interested in running the TV Kitchen as a developer please review our [install instructions](docs/INSTALL.md).
+If you are interested in running a TV please review our [install instructions](docs/INSTALL.md).
```
reefdog, line 11: This seems garbled. "running a TV" doesn't seem to be what you mean to say!

package.json
```diff
@@ -25,7 +25,7 @@
     "journalism",
     "news"
   ],
-  "author": "Bad Idea Factory <author@biffud.com>",
+  "author": "Bad Idea Factory <biffuddotcom@biffud.com>",
```
reefdog, line 28: This email address is a crime, but a delicious crime, like when all that Pappy Van Winkle went missing.

README.md
slifty: image

package.json
slifty: I am genuinely assuming this is [a Wishbone reference](https://www.youtube.com/playlist?list=PLDymFoFsHNt14kLH5F3tGy9x5-579qwUq).

reefdog: I just stumbled across this while looking for something else. This is what I get for ignoring all the GitHub notification emails.

No, it was a reference to [this](https://www.huffpost.com/entry/pappy-van-winkle-stolen_n_4114677).

PR #90, opened 2020-08-27
Remove module-alias package

### LINKED ISSUE #89: Getting rid of module-alias
#### Description
Now that countertop is going to be a package, we need to remove `module_alias`.  This was identified a while ago by @chriszs in #51.

To quote Chris:

------

`module-alias` theoretically supports imports, and does so under Babel, but the way it does so is to hijack every subsequent `require`. ~~That seems to mean I need to require it? (E.g. `require('module-alias/register')`).~~ (Correction: it's the `/register` part it doesn't seem to like. `import register from 'module-alias'` might work?) And it _seems_ ineffective at rewriting imports if they don't go through Babel.

One pitfall of the require hijack approach (internally it uses a require hijacker called `pirates`, amusingly, according to my stack traces) is acknowledged in its README: 
>WARNING: This module should not be used in other npm modules since it modifies the default require behavior! It is designed to be used for development of final projects i.e. web-sites, applications etc.

This could be a barrier to adaptive reuse, even if the core TVKitchen module is primarily for running directly.

There are alternative approaches. One [detailed here](https://arunmichaeldsouza.com/blog/aliasing-module-paths-in-node-js) uses `link-module-alias`, which creates symlinks as a post-install step run after every `npm install` or `yarn`. That post also mentions yarn workspaces as an alternative, one already being used in this project elsewhere.

One benefit is those approaches _might_ need fewer shims, e.g. for ESLint and Jest. But there may be other tradeoffs.

------

* (https://arunmichaeldsouza.com/blog/aliasing-module-paths-in-node-js

#### Related Issues
None

slifty:
## Description
This PR removes `module-alias`, replacing all of those aliased imports with relative paths.  It's a bit more cluttered and depressing, but it had to happen in order to support npm packages.

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits
- [x] Changelog(s) updated

## Steps to Test
1. `yarn test`

## Deploy Notes
- `yarn install`

## Related Issues
Resolves #89


src/scripts/_sandbox.js
```diff
@@ -4,8 +4,3 @@
 //
 // Do not commit changes to this file! In fact, you should locally set the `--assume-unchanged`
 // flag on it: `git update-index --assume-unchanged ./src/scripts/_sandbox.js`
-import 'module-alias/register'
-
-/**
- * DO NOT MODIFY ABOVE THIS BLOCK
- */
```
reefdog, line 11: Did you mean to remove the "DO NOT MODIFY ABOVE THIS BLOCK" comment?

slifty: I did -- it doesn't matter what people do to that file since it (shouldn't be) tracked.

It used to be the case that we needed that import for anything to work correctly from sandbox.

PR #94, opened 2020-08-28
Add injectable logging

### LINKED ISSUE #91: Make logging customizable
#### Description
Now that this is an npm package (as opposed to a standalone application), it does not make as much sense for it to have `.env` driven logging.

We should instead look into exposing an API that lets a user of the package define a custom logger, similar to what [kafkajs does](https://kafka.js.org/docs/custom-logger).

The log levels are already exposed in `@tvkitchen/base-constants`, and the countertop elements already take in settings on construction, so we're actually partly there.

What is less clear to me is if we should remove winston from the project, and if so, should that be part of this issue.  Use of winston as the default might be a little opinionated / overkill as opposed to some kind of trivial console based logger default.

- https://kafka.js.org/docs/custom-logger

#### Related Issues
None.

slifty:
## Description
This PR adds the ability to specify a logger when creating a new `Countertop`.

This PR removes reliance on `winston`.

This PR adds a default logger that uses `console.log`.

As part of the addition of inject-able logging, this changes the way appliance settings are passed to the countertop.

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits
- [x] Changelog(s) updated

## Steps to Test
1. `yarn test`

## Deploy Notes
- `yarn install`

## Related Issues
Resolves #91


src/classes/Countertop.js
```diff
@@ -9,20 +10,36 @@ import CountertopTopology from './CountertopTopology'
...
-	 * @param  {Class}  Appliance     The class of the IAppliance being registered.
-	 * @param  {String} configuration The configuration used to register the appliance station.
-	 * @return {CountertopStation}    The station created for this appliance.
+	 * @param  {Class}  Appliance                 The class of the IAppliance being registered.
+	 * @param  {String} options.applianceSettings The settings to be passed to the appliance.
+	 * @return {CountertopStation}                The station created for this appliance.
 	 */
-	addAppliance = (Appliance, configuration) => {
+	addAppliance = (Appliance, {
+		applianceSettings = {},
```
slifty, line 35: This might feel like overkill to use destructuring for.  The thinking is that there are going to be additional settings that are more specific to countertop topologies (e.g. "ignoreStations" etc.)

src/tools/loggers/consoleLogger.js
```diff
@@ -0,0 +1,11 @@
+import { logLevels } from '@tvkitchen/base-constants'
+
+export default {
```
chriszs, line 3: Some thoughts:

1. This feels a little like a DRY violation, but maybe not one that would be better if it weren't.
2. Do we want to output errors to the error output stream?
3. We're depending on the context this is called in to define the logger in a property called "logger." Are we sure we want to couple it that way?
4. Most of this is defined by the built-in logger, except `fatal` and prepending the level to the message. Theoretically we could just pass in `console` except for those requirements.

src/classes/Countertop.js
chriszs: It feels a little aesthetically weird to imagine calling `addAppliance` like:
```js
addAppliance(appliance, { applianceSettings: { setting: false } })
```

But not _that_ weird.

What do you mean by "overkill?"

chriszs:
Presumably [the logger in IAppliance](https://github.com/tvkitchen/base/blob/e71d4493606fd2acac61251c84971d1c259114cc/packages/interfaces/src/IAppliance.js#L15-L23) goes away after this lands, right?

src/classes/Countertop.js
slifty: I just meant that without other parameters being destructured it seems silly to destructure parameters.

That being said... you're absolutely right.  I'm going to take this back.  Maybe we should just have three parameters:

1 -> the appliance
2 -> appliance settings
3 -> whatever relevant countertop settings we end up deciding on.

Where 3 is the one that might be destructured once we decide on those settings.

src/tools/loggers/consoleLogger.js
slifty: 1. which part?
2. Oh good thought (also, good thought re #4...)
3. Nope that is totally a mistake!  Also I should have tests which would have exposed that mistake.
4. Yes!  I'll go do that.

slifty:
Regarding the IAppliance logger -- different repositories so they're independent; we could consider making the ConsoleLogger something added to `base-classes` which gets imported by both contexts.

Also, I think that IAppliance probably shouldn't have an implemented logger anyway, and the default logger should actually be part of AbstractAppliance.

src/classes/Countertop.js
slifty: I just meant that without other parameters being destructured it seems silly to destructure parameters.

That being said... you're absolutely right. I'm going to take this back. Maybe we should just have three parameters:

1 -> the appliance
2 -> appliance settings
3 -> whatever relevant countertop settings we end up deciding on.

Where 3 is the one that might be destructured once we decide on those settings.

src/tools/loggers/consoleLogger.js
slifty: OK I grok the DRY comment now.

Worth noting that if we just took `FATAL` out of `logLevels` we would be able to just use `console` as a default logger without needing a logger class anywhere...

The difference between `FATAL` and `ERROR` is the scope of the error.  If it takes down the whole service, it causes FATAL; if it's just taking down the local operation, it's ERROR...

Not sure we'd be giving up much by dropping fatal tbh.

slifty: Redid this with a little inspiration from how [kafkajs](https://github.com/tulios/kafkajs/blob/master/src/loggers/console.js) does it -- I think I avoided a bit of the DRY smells but I imagine it isn't perfect.

Overall I am comfortable with the idea that we will probably want to revisit this -- both to add better structure, probably to remove `fatal` and the like as well depending on how the [conversation in base-constants goes](https://github.com/tvkitchen/base/issues/84).

src/tools/loggers/consoleLogger.js
```diff
@@ -0,0 +1,35 @@
+/* eslint-disable no-console */
+// Since using `console` here is kind of the whole point.
+import { logLevels } from '@tvkitchen/base-constants'
+
+const log = (level, message) => {
+	const structuredMessage = JSON.stringify({
+		level,
+		message,
+	})
+	switch (level) {
```
chriszs, line 10: Another way to do this would be
```js
console[level === 'fatal' ? 'error' : level](structuredMessage)
```

src/tools/loggers/consoleLogger.js
```diff
@@ -0,0 +1,35 @@
...
+		case logLevels.trace:
+			return console.trace(structuredMessage)
+		default:
+			return console.log(structuredMessage)
+	}
+}
+
+export default {
+	log,
+	fatal: (message) => log(logLevels.fatal, message),
```
chriszs, line 29: Could also do:
```js
fatal: log.bind(this, logLevels.fatal),
```

src/tools/loggers/consoleLogger.js
slifty: I think the "tradeoff" is the loss of default behavior (e.g. if someone ran `logger.log("PANTSPARTY", "wooooooo!"` it would error)...  but that's probably OK?

slifty: Wait... I have an idea...

slifty: Oh the other thing that shorthand approach does is rely on the (coincidental?) coupling of the log level constants being identifcal to the console method names (e.g. lower case).

I agree it feels verbose but I think that for now I'll just leave it as is and honestly we'll probably wind up deleting this whole thing later when we revisit logging...

slifty:
OK so I think this is good enough for now -- I hear you on the removing lines from the `consoleLogger.js` but there is enough reason to keep it that way for now that I will keep it there, and I expect it to become cleanup later.

The more long-term-important suggestions around things like API shape have been made so I think this is good to merge in!

src/tools/loggers/consoleLogger.js
slifty: How does bind even work 😬

chriszs:
Yeah, I pretty much understood you were going to keep this design when I did my review. I think it's okay.

PR #98, opened 2020-09-06
Create countertop workers

### LINKED ISSUE #84: Allow the countertop to start processing data
#### Description
We currently have a Countertop that can accept new appliances and use that set of appliances to generate a stream topology.

The next step is to actually use that topology to create workers and start generating / processing data as per a given appliance configuration.

Using the API defined in #82, enable the `start` and `stop` methods for the `CountertopCoordinator`.

This is going to involve creating the concept of a `CountertopWorker` and thinking carefully about how to differentiate streams within Kafka.

None

#### Related Issues
- #82 
- #23

slifty:
## Description
This PR adds the concept of CountertopWorkers, which are the portion of the countertop to actually interact with Kafka and invoke appliances.

This PR also adds unit tests for all countertop components.

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits
- [x] Changelog(s) updated

## Steps to Test
0. `yarn install`
1. `yarn test`

## Deploy Notes
- `yarn install`

## Related Issues
Resolves #84


src/classes/Countertop.js
```diff
@@ -43,8 +44,48 @@ class Countertop {
...
+	 * with those stations.
+	 *
+	 * Running this is destructive: it will stop processing, remove all previous streams from
+	 * the countertop and replace them with updated streams.
+	 *
+	 * @return {CountertopTopology} The new countertop topology.
+	 */
+	updateTopology = async () => {
+		this.logger.debug('CountertopCoordinator: updateTopology()')
+		await this.stop()
```
slifty, line 84: I think instead of invoking stop, updateTopology should verify that internal state preconditions are met and fail if they are not.

For one thing, this would mean updateTopology can be synchronous, but more importantly it would make the impact of calling it much more predictable.

src/classes/Countertop.js
```diff
@@ -43,8 +44,48 @@ class Countertop {
...
+	/**
+	 * Start the countertop and begin data processing.
+	 *
+	 * This will regenerate the countertop topology to reflect all currently added appliances.
+	 *
+	 * @return {Boolean} Whether the countertop successfully started.
+	 */
+	start = async () => {
+		this.logger.debug('CountertopCoordinator: start()')
+		await this.updateTopology()
```
slifty, line 56: We should move updateTopology to the `addAppliance` -- that way the topology is always assumed to reflect the appliances added.

src/classes/Countertop.js
```diff
@@ -14,6 +15,8 @@ class Countertop {
 
 	stations = []
 
+	state = countertopStates.STOPPED
```
slifty, line 18: I forget if we decided it's bad / error prone to set values here vs in the constructor

src/classes/CountertopWorker.js
```diff
@@ -0,0 +1,132 @@
...
+	 *
+	 * @param  {Class}            Appliance     The class for the appliance that this worker will use.
+	 * @param  {CountertopStream} stream        A stream that ends in this worker's station.
+	 * @param  {Object}           configuration Configuration settings for the Appliance.
+	 */
+	constructor(
+		Appliance,
+		applianceSettings = {},
+		{
+			stream,
```
slifty, line 39: Is passing stream here appropriate?

It is used to decide which kafka topics to track / what the kafka group should be.  Note: nobody knows that workers use Kafka this way, which is good (vs, for instance, passing in things like `kafkaStream` or `kafkaGroup`)

Tl;DR it feels odd for the Worker to have to know its context, but ultimately it does need to know what data streams to read and write from, and all of that is contained by the CountertopStream that is passed here.

src/classes/CountertopWorker.js
```diff
@@ -0,0 +1,132 @@
...
+			logger = consoleLogger,
+		} = {},
+	) {
+		this.stream = stream
+		this.logger = logger
+		this.consumer = kafka.consumer({ groupId: stream.id })
```
slifty, line 45: Should we give workers their own UUIDs (rather than relying on the stream UUID)

There should only ever be one worker per stream, so it would be 1:1, but still.

src/classes/CountertopWorker.js
```diff
@@ -0,0 +1,132 @@
...
+		if (!await this.appliance.audit()) {
+			return false
+		}
+		if (!this.appliance.start()) {
+			return false
+		}
+		await this.admin.connect()
+		await this.producer.connect()
+		await this.consumer.connect()
+		const inputTopics = this.appliance.constructor.getInputTypes().map(
```
slifty, line 80: Maybe should be a named utility

src/classes/Countertop.js
```diff
@@ -40,13 +48,69 @@ class Countertop {
...
+	 *
+	 * This will regenerate the countertop topology to reflect all currently added appliances.
+	 *
+	 * @return {Boolean} Whether the countertop successfully started.
+	 */
+	start = async () => {
+		this.logger.trace('CountertopCoordinator: start()')
+		this.state = countertopStates.STARTING
+		const promises = this.stations.map(async (station) => station.start())
+		if ((await Promise.all(promises)).every((result) => result)) {
```
chriszs, line 66: Okay, so `station.start()` returns a promise with a boolean indicating success or failure? Failure isn't indicated by a promise rejection? I could go look, I guess, but I'm just thinking this one through out loud instead.

src/classes/Countertop.js
```diff
@@ -40,13 +48,69 @@ class Countertop {
...
+	 * Stop the countertop and halt all data processing.
+	 *
+	 * @return {Boolean} Whether the countertop successfully stopped.
+	 */
+	stop = async () => {
+		this.logger.trace('CountertopCoordinator: stop()')
+		this.state = countertopStates.STOPPING
+		this.stations.map((station) => station.stop())
+		const promises = this.stations.map(async (station) => station.stop())
+		if ((await Promise.all(promises)).every((result) => result)) {
```
chriszs, line 84: I wonder if instead of jamming the promise one-liner into this if statement, this wouldn't be cleaner as a variable.

src/classes/CountertopWorker.js
chriszs: Seems fine.

src/classes/CountertopWorker.js
chriszs: Seems like a good idea. If the constraint you describe (and which I still don't fully understand) holds it's unneeded, but less coupling seems good.

src/classes/CountertopWorker.js
```diff
@@ -0,0 +1,132 @@
...
+			topics: outputTopics.map(
+				(topic) => ({
+					topic,
+				}),
+			),
+		})
+		await Promise.all(
+			inputTopics.map(
+				async (topic) => {
+					await this.consumer.subscribe({ topic })
```
chriszs, line 103: Not sure you want to async/await here, since you're doing that with `Promise.all`.

src/classes/CountertopWorker.js
```diff
@@ -0,0 +1,132 @@
...
+		)
+		const outputTopics = this.appliance.constructor.getOutputTypes().map(
+			(outputType) => getStreamTopic(
+				outputType,
+				this.stream,
+			),
+		)
+		await this.admin.createTopics({
+			waitForLeaders: true,
+			topics: outputTopics.map(
```
chriszs, line 94: I wonder if the generation of the list of TopicConfigs from topic strings could be broken out into a variable and go outside the function call.

chriszs (changes requested):
Hey there, just a couple stylistic suggestions and questions about how we're using promises.

src/classes/Countertop.js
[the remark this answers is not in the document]
slifty: Yeah that's the case (start is documented as returning true on successful start, false otherwise).

src/classes/Countertop.js
```diff
@@ -40,13 +48,69 @@ class Countertop {
...
+	 * @return {Boolean} Whether the countertop successfully started.
+	 */
+	start = async () => {
+		this.logger.trace('CountertopCoordinator: start()')
+		this.state = countertopStates.STARTING
+		const promises = this.stations.map(async (station) => station.start())
+		if ((await Promise.all(promises)).every((result) => result)) {
+			this.state = countertopStates.STARTED
+		} else {
+			this.state = countertopStates.ERRORED
```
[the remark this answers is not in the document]
slifty: OK so I actually went back and forth and wonder the same thing.

I originally had a `setState` method -- but then realized I don't actually want anything outside of this instance to modify the state, and I was worried that `setState` existing would indicate that it's "OK" to call.

e.g. I never want:

```
const countertop = new Countertop()
countertop.setState('WHEEEEEEEEEEEEeeeeeeeeeeeeeeeeee')
```

What do you think.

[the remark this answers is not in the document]
slifty: I guess it should probably just be a method that is not documented as being part of the 'public' API.

src/classes/CountertopWorker.js
slifty: Deciding to not bike shed this class yet since I expect it to be in a bit of flux.  I think both `start` and `stop` are a doing a bit too much without relying on utility methods.

PR #102, opened 2020-09-07
Make kafka configurable via API

### LINKED ISSUE #99: Make kafka configurable
#### Description
Similar to our removal of logging from the environment configuration (#91) we should not rely on environment variables for kafka.  Instead, necessary kafka settings should be passed to the countertop directly upon instance creation.

I expect the list of configurable settings to expand over time, but we can start with the two fields actually used:

- clientId: 'tv-kitchen',
- brokers: [config.KAFKA_BROKERS],

None

#### Related Issues
- #69

slifty:
## Description
This PR adds the ability for a developer to specify kafka configuration settings when creating a Countertop.  In turn, it removes the (very simple) `kafka` tool we created and the ability to configure kafka via environment variables.

Kafka injection is now a part of the core countertop components (countertop => stations => workers)

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits
- [x] Changelog(s) updated

## Steps to Test
1. `yarn test`

## Deploy Notes
- `yarn install`

## Related Issues
Resolves #99

### DIFF
--- a/src/classes/Countertop.js
+++ b/src/classes/Countertop.js
@@ -1,3 +1,4 @@
+import { Kafka } from 'kafkajs'
 import { countertopStates } from '../constants'
 import { consoleLogger } from '../tools/loggers'
 import CountertopStation from './CountertopStation'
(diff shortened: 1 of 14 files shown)


src/classes/Countertop.js
```diff
@@ -13,18 +14,26 @@ import CountertopTopology from './CountertopTopology'
...
 	/**
-	 * @param  {Logger} options.logger A logger with methods for all TV Kitchen logLevels.
+	 * @param  {Logger} options.logger        A logger with methods for all TV Kitchen logLevels.
+	 * @param  {Object} options.kafkaSettings Kafka settings as defined in the kafkajs library.
 	 */
 	constructor({
 		logger = consoleLogger,
+		kafkaSettings = {
+			clientId: 'tv-kitchen',
+			brokers: ['127.0.0.1:9092'],
```
slifty, line 31: I'm not sure we should actually provide any default values at all, for anything here.

slifty: I'm going to take out this default for now.  We can add one some day if we want to.

PR #111, opened 2020-09-20
Fix message ordering

### LINKED ISSUE #110: Writing to kafka isn't guaranteed order
#### Current Behavior
The CountertopWorker does not await the response from Kafka when writing, which means that it is possible that a given message will not be written in order.

This bug may sound familiar, since we addressed it when we did ingestion engines originally (but I forgot when writing the countertop worker).

#### Input

N/A -- this is just a bug in the business logic generally, so it would potentially apply to any topology.

#### Expected Behavior

Kafka messages should always be written in the order that payloads are emitted by an appliance.

#### Possible Solutions

Instead of writing directly to kafka, the payload event handler should write to a Writable stream.  That stream should then honor order.

slifty:
## Description
This PR fixes CountertopWorker so that it will not write to Kafka until the previous payload has been processed. 

## Due Diligence Checklist
- [x] Tests
- [x] Documentation
- [x] Consolidated commits
- [x] Changelog(s) updated

## Steps to Test
1. `yarn test`

## Deploy Notes
None

## Related Issues
Resolves #110

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,10 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Fixed
+- Message ordering from CountertopWorkers are now guaranteed.
 
 ## [0.1.0] - 2020-09-09
-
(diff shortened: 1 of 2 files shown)


slifty:
Merging this without approval despite the critical bug it introduces.  Also creating an issue to reflect that memory leak.

Issue #153, opened 2022-05-12
Move to Typescript


slifty:
**Feature description**

This project would make more sense as a Typescript project; there is already (awkward) type checking going on.

It will be much easier to write appliances with TS because of the hinting and tooling.

Countertop is a fine place to start this conversion.

slifty:
This is going to be handled in stages.  For instance:

1) Enable TypeScript alongside the current build patterns.
2) Begin to transfer some of the types defined in https://github.com/tvkitchen/base/ as TypeScript
3) Begin to convert some of this repository in independent chunks.
4) Once everything is in TS, disable the babel and move to pure TypeScript.

PR #155, opened 2022-05-13
Support TypeScript

slifty:
This PR adds support for TypeScript to the respository.

Note that it this is simply the first step in the process of converting this repository to TypeScript.  The long term goal is to be able to remove Babel (and JS) entirely.

Related to #153


.eslintrc.typescript.json
```diff
@@ -0,0 +1,32 @@
+{
+		"parser": "@typescript-eslint/parser",
```
slifty, line 2: Indentation is off here.

Issue #154, opened 2022-05-13
Base should be baked into countertop


slifty:
Especially once #153 is complete I think it will make a lot of sense for all of the "base" and "core" functionality to get merged into `countertop`

This means the flow for a developer would be to import countertop and do whatever they want to do without needing to worry about importing a bunch of random "base-error" or "base-interface" junk.  Countertop will truly be the entry point not just for appliance users but for appliance developers.  This will also make it easier to make sure a countertop + appliance pair is in sync in terms of what version of constants exist instead of updating a bunch of redundant dependencies.

slifty:
I don't think there's any good reason to import the `errors` package / classes.  Most of those were just there to support... type checks.

PR #160, opened 2022-06-18
Add Payload class + types

slifty:
This PR adds the Payload class along with relevant typescript objects.  It also sets up `jest` to be able to run against TypeScript.

I've decided that documentation for things like `Payload` should live as part of the [tv.kitchen](https://tv.kitchen) website.  Having it also in the repository makes it hard to know where to look, and I don't want to have too many different places for documentation.

Related to #154 
Related to #153


slifty:
I think when upgrading jest I ... upgraded basically all of our packages.  So long as tests pass I think that's OK, even if an ideal world would have that happen in a separate PR.

slifty:
Update: I undid that accidental universal upgrade; I'll do that another time.

PR #164, opened 2022-06-20
Support multiple coverage reports

### LINKED ISSUE #161: CodeCov is off
1. The badge is pointing to the wrong branch (`master` rather than `main`)
2. When I added a second "run" of jest for TypeScript it meant that ONLY the TypeScript test coverage is being reported to CodeCov.  This is... probably ok really since most new code will be TS based.

I'm wary of spending time fixing the second issue since I'd rather spend that time just moving over to TypeScript and making the issue moot.  Still, it's captured here in case I want to take it on.

slifty:
This PR fixes our codecov analysis pipeline to process TS and JS separately, rather than ignoring JS.

Resolves #161

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -17,5 +17,11 @@ jobs:
         node-version: 16.x
     - run: yarn install
     - run: yarn build
-    - run: yarn test --coverage
+    - run: yarn test:js --coverage
     - uses: codecov/codecov-action@v2
+      with:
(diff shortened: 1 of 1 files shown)


slifty:
I believe CodeCov won't actually comment here because the only file changed is untracked (e.g. the workflow).

That said, when I log into codecov it does appear to report on BOTH the TS and JS files

image

Merging!

PR #171, opened 2022-07-11
Fix test coverage and linting

slifty:
This PR fixes our test coverage reporting to include files that don't have tests.  It also makes sure we're explicitly using `airbnb-base` linting rules.


slifty:
This makes it look like coverage went down (by 1%) but really this is just recognizing a more accurate reality because it ensures all testable files are recognized in the test coverage report.

PR #186, opened 2023-07-21
Remove babel from build

slifty:
This PR removes babel from the lint and build steps and replaces it with tsc / ts-lint.

Unfortunately for jest this is a bit awkward, since I couldn't really figure out how to get jest to handle both typescript AND es6 module syntax at the same time.  For now I had to just modify tests to directly import js files rather than touch indexes that import ts files (which caused jest to get upset about typescript syntax in a js context)

Related to #153


slifty:
Closing this in favor of #187

PR #187, opened 2023-07-22
Configure babel to handle typescript

slifty:
This PR modifies our build and lint steps (and I suppose test step) so that babel can understand typescript files that have been referenced by javascript files.

This PR is a replacement for https://github.com/tvkitchen/countertop/pull/186, which tried to go in the other direction (elimintating babel entirely and having typescript handle javascript).  I think this is a cleaner approach for now due to the fact that jest needs babel in order to handle ESM.

Related to Issue #153


slifty:
Wanted to note that this approach was inspired by @louh via https://github.com/streetmix/streetmix/commit/05feb8b9cf8eff6938397d12bf09e99477f0607b

louh:
Glad it worked!

PR #188, opened 2023-07-22
Move CountertopWorker to TypeScript

slifty:
This PR moves the CountertopWorker class to TypeScript.

It also brings the AbstractAppliance from our [@tvkitchen/base](https://github.com/tvkitchen/base/blob/main/packages/interfaces/src/IAppliance.js) over to this project (and converts it to TypeScript as well)

Related to #153 
Related to #154


slifty:
cc @louh this PR actually represents the migration of a js class to typescript (check out the final commit for the meat and potatoes)

slifty:
So here's my thinking:

1. `CountertopWorker` did not have meaningful test coverage in the first place
2. A lot of `CountertopWorker` is based in kafka interaction, and I do not think we want to keep kafka as part of TVK in the long run.

Which is to say... I don't think I'm going to use this PR as an opportunity to address the poor test coverage even though codecov is upset about it.

PR #191, opened 2023-07-26
Convert CountertopStation to TypeScript

slifty:
This PR converts CountertopStation to TypeScript

Related to #153


slifty:
I think this is another case where test coverage is going to be too hard to do since we rely on Kafka and aren't mocking kafka effectively.  Once Kafka is taken out of the project it will be easier to test the rest of CountertopStation.

